PR #5, opened 2018-01-08
Remove remnants of Python from package.json

azz:


package.json
```diff
@@ -1,14 +1,13 @@
 {
-  "name": "@prettier/python",
+  "name": "prettier-plugin-python",
```
j-f1: `prettier-plugin-java`?

azz: d'oh!

thorbenvh8:
@j-f1 thanks for finding the bug :)
@azz thanks for the pr :)

azz:
It's 4am. I need sleep. 💤

PR #87, opened 2018-11-08
Project Restructuring

bd82:
* Refactor project to a mono repository structure with 
 2 packages, one for the parser and one for the prettier plugin itself.

* Refactor tests to use Mocha + Chai instead of Jest.
  This was done because Jest runs each test in a separate process
  Thus we pay the Parser init cost many many times, e.g the parser's test
  suite went from 14 seconds to 200ms on my machine...

* Added a script for massive inputs sanity tests that clones
  java OSS repositories and attempts to parse all their java files.


bd82:
Sorry for too many changes at once, there was no proper upstream repo to work with before the migration.

jdubois:
Oh indeed that's a huge change! Let's do this first, and work on the license stuff just afterwards.

I'm putting @thorbenvh8 as a reviewer - indeed, I think that should be fine for him, but as he originally expressed concerns on this, I'd rather have his opinion.

-> @thorbenvh8 as I know you don't have much time for all this, and I expect you trust us, I'll merge this in one week if you don't add any comment

bd82:
The intention in switching this to a multi package repository (a.k.a mono repo) is to speed up development velocity as all the parts are in the same repo so there is no need to manage multiple versions and releases.

To read more about mono repos, visit: https://github.com/babel/babel/blob/master/doc/design/monorepo.md

Note that this does not conflict with having the Java-Parser (formerly cherotain-java) package consumed separately and re-used in other projects. As a mono-repo will release **multiple** packages to npm.
* Optionally even with separate version numbers.

I even think the repository could have **three** distinct packages, (parser / AST builder / re-writer).
So it is not about creating a monolith, rather a different way to organize the modules.

In the long term a re-separation can be re-evaluated (easy to implement), but in the short term
I believe it would speed up the development velocity, which is normally very important, but even more so now as our development resources are somewhat constrained...

bd82:
I think I will merge this to a separate branch so I can continue adding **incremental** changes 
which otherwise will get lost in the huge commit.

bd82:
I merged this into a side branch. So I can continue working on the upstream project instead of a fork.
We can leave this PR open to get approved by @thorbenvh8 in the meanwhile.

PierreBesson:
I think @bd82, you knows what he is doing. You should lead the work for now has you have a lot of experience in this domain. We will try to find interns at Ippon to help.

On a side note, I would love to have a monorepo structure for other JHipster project (jhipster-framework and generator-jhipster) but everyone is against me on this 😛.

bd82:
> You should lead the work for now

I am trying to give a boost to this project in the short term, but keep in mind I am not planning on a long term constant involvement.  So I am focusing on making things as easy as can be to develop / contribute to for now.

> I would love to have a monorepo structure for other JHipster project (jhipster-framework and generator-jhipster) but everyone is against me on this.

Mono repos have a name with some very bad associations 😄 (**Mono**Lith).
Developers tend to want to get rid of their monoliths not build new one.

deepu105:
Btw do you think doing this in Typescript will give benefits overall?

murdos:
I'm a bit worried that we'll loose commit history for most of the files.
I've a script that I've already used in the past that combine 2 different git repo in a unified new repository, each previous repo becoming a subdirectory in the new repository, and preserving history.
I can do the same operation here if needed.

bd82:
> Btw do you think doing this in Typescript will give benefits overall?

### TLDR
Too few benefits relative to the costs for this **specific** project.

### In Depth Reasoning.
I've got some experience with TypeScript projects, 
There is an **overhead** for introducing TypeScript, which I prefer not to pay it for small projects.

This project is not small, However I think that TypeScript **won't be a great benefit** to this project because:
- The parser code is very structured and uses a limited set of APIs, so the benefits of type checks are smaller.
- The visitor code cannot benefit from TypeScript because the structure it traverses (CST)
  is dynamic by nature and has no signatures.
- The prettier re-writer works on the AST created by the visitor, so it could certainly benefit from type definitions to the AST, however the re-writer quality can be assured by other (stronger) means such as:
  * High code coverage.
  * **Cyclic** tests that compares the AST of a java file before and after formatting.
- The public API of this project is tiny.

The real benefit of TypeScript would be to promote re-use of the java-parser
package as a standalone package (outside prettier context), by providing and validating
against TypeScript interfaces for the AST.

bd82:
> I'm a bit worried that we'll loose commit history for most of the files.
I've a script that I've already used in the past that combine 2 different git repo in a unified new repository, each previous repo becoming a subdirectory in the new repository, and preserving history.
I can do the same operation here if needed.

I must admit did not consider the history perspective.
I am worried that if we use such a script I would have to re-do a-lot of the 
work involved in the restructuring. is that avoidable?

jdubois:
@bd82 you're clearly the best person here for doing such a refactoring, so please go ahead. When you find that this is good, unless @thorbenvh8 doesn't agree, don't hesitate to merge.

To me, we will then have few things to update, but that should be separate PRs we do afterwards:

- Change the license to Apache 2, so the project has the same license as the other JHipster projects. Also add the license headers and notice file. Let me do this, I'm used to that.
- Add our code of conduct (I'll also do this)
- Migrate from Yarn to NPM (as all other JHipster projects use NPM)
- Have a look at the CI system, and migrate to Travis and/or Azure devops pipelines if possible ( @pascalgrimaud will be the best here)
- Probably work on the README and the contributing guide, and maybe even create a specific section on www.jhipster.tech - the goal is to attract more people to the project

pascalgrimaud:
@bd82 : yes, don't hesitate to ping me if you want me to work on the CI

murdos:
> I am worried that if we use such a script I would have to re-do a-lot of the
> work involved in the restructuring. is that avoidable?
I don't think that it's possible. At least the big restructuring commit should be redone.
I could have helped on this task but it seems I'm the only one worried, so I guess you should go ahead.

bd82:
merged in #91 I will open issues for other topics raised here.

bd82:
@murdos It seems lerna devs already considered this issue.
See: https://github.com/lerna/lerna/tree/master/commands/import

Now we know for any future mono repos 😄

PR #91, opened 2018-11-09
Mono repo restructure

bd82:
Previous discussion in #87 

This includes additional tests structure updates, the beginning of a contribution guide and others


bd82:
I am merging this as otherwise with such a large change  in progress other work gets **blocked**.
I worked with @thorbenvh8 on chevrotain-java (parser package) and I am aware he wanted the parser
to be consumable as a separate package(I want that too). This mono-repo structure **does not** prevent that. Nor does it prevent a separate version life cycle, or even reverting to separate repos (easily) in the future.

So lets start working and resolve any (potential) concerns raised by @thorbenvh8 in an **asynchronous** manner. 😄

Issue #92, opened 2018-11-09
CI status checker


bd82:
- @jdubois wrote:
 > Have a look at the CI system, and migrate to Travis and/or Azure devops pipelines if possible ( @pascalgrimaud will be the best here)

Questions:
- Any specific reason circleci cannot be used? it is a much more powerful and customizable build system
  is it because you want a consistent build system among all the JHipster projects?

I can use either (although a circleci configuration file is already present) just wondering...

bd82:
With CircleCI it is possible to use  custom docker images and also to define more frequent cron jobs.
But considering it is a simple (CI wise) project it does not truly matter.

Can you please enable travis @pascalgrimaud and I will create the travis.yml file?

pascalgrimaud:
did you do something ?
I already saw: https://travis-ci.org/jhipster/prettier-java

bd82:
I don't have permissions to activate it.

Maybe it was already configured due to the repository migration and got triggered once it saw a travis.yml file?

pascalgrimaud:
I don't have permission either.
Only Julien and Deepu can. But I supposed it activated automatically so it's cool. 

Don't hesitate to ping me or assign me the ticket, if you need help on CI.

bd82:
> Don't hesitate to ping me or assign me the ticket, if you need help on CI.

Thanks @pascalgrimaud 

Everything seems to be working now. 😄 

May need help at a later stage if/when we want to automate the release process.

pascalgrimaud:
There is no automate release process today, and it is something I would like to do.
So it will be cool to start at least one, and why not for this project!

bd82:
It is fairly simple for regular npm packages.
See how I do it in another project.
- [Local scripts](https://github.com/bd82/regexp-to-ast/tree/master/scripts) that update changelog date, package.json version and push a new  **git tag**.
- [Trigger in circle-ci](https://github.com/bd82/regexp-to-ast/blob/master/.circleci/config.yml#L49-L54) to perform deploy job on **tag push**.
- Use CI **secure** env variable to authenticate with **npm token**
  and run **npm publish**
  * https://github.com/bd82/regexp-to-ast/blob/master/.circleci/config.yml#L39-L40

prettier-java however uses lerna, and I'm not sure what add complexity it may add as its the first time I am using it myself.. 😄

jdubois:
For the Travis build: it was automatically set up in Travis but I did check it.

PR #95, opened 2018-11-09
Travis build configuration.

bd82:
Part of #92


bd82:
that is not the source code i meant to push.

PR #120, opened 2019-01-10
License and Notice file updates.

bd82:
- MIT --> Apache.
- I tried to use the same style as in other JHipster org projects.


bd82:
I think @thorbenvh8 must approve this as even after the re-write there is still some code here
that he originally wrote, so a license change must be approved by him.

bd82:
I am merging this so at least the license the java-parser package is released
on would be correct.

@jdubois please let me know if any terminology changes are needed...

Btw is the JHipster organization a **legal entity**? 
The copyrights terminology in some JSFoundation projects seems (to me) the best:
- https://github.com/lodash/lodash/blob/master/LICENSE#L3
- https://github.com/mochajs/mocha/blob/master/LICENSE#L3

j-f1:
Here’s what we do for Prettier core:

https://github.com/prettier/prettier/blob/282597c1e5583d1c8682af8d947795f09efeffd5/LICENSE#L1

jdubois:
@bd82 also for JHipster we always add a "NOTICE" file, as by the license people should copy it in their project. This is important as JHipster is being copied in a few products, not always with good intentions, and that adds another layer of protection.

bd82:
There already is a NOTICE file in this commit but I think it has some copy pasta mistakes.

@j-f1 Thanks for the reference.
I copied the terminology from generator-jhipster, it seems there may be a few variants of the terminology present inside the Jhipster org...

jdubois:
Thanks @bd82 - OK I will reword it a little bit

bd82:
> Thanks @bd82 - OK I will reword it a little bit

Great 😄

PR #128, opened 2019-01-22
Guava failures

FauxFaux:
The library currently fails to parse 101 files inside Guava.

This PR currently adds a (failing) test.

I'll.. raise individual issues for the failing bits of syntax?

### DIFF
--- a/packages/java-parser/package.json
+++ b/packages/java-parser/package.json
@@ -12,5 +12,14 @@
     "test": "npm-run-all clone-samples test:execute",
     "test:execute": "mocha \"test/**/*-spec.js\"",
     "clone-samples": "node ./scripts/clone-samples"
+  },
+  "devDependencies": {
+    "chai": "^4.2.0",
+    "eslint": "^5.12.1",
+    "fs-extra": "^7.0.1",
+    "klaw-sync": "^6.0.0",
+    "lodash": "^4.17.11",
+    "mocha": "^5.2.0",
+    "npm-run-all": "^4.1.5"
   }
 }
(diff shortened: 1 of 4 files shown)


FauxFaux:
I found at least #129, #130, and #131.

bd82:
Thank for reporting this and for the issues 👍

packages/java-parser/package.json
```diff
@@ -12,5 +12,14 @@
     "test": "npm-run-all clone-samples test:execute",
     "test:execute": "mocha \"test/**/*-spec.js\"",
     "clone-samples": "node ./scripts/clone-samples"
+  },
+  "devDependencies": {
```
bd82: I am not sure we want these common dev deps at the package level as it would create
duplicates and **multiple** versions of dev tools co-existing at the same time on the same mono repo.

How do other mono repos handle this issue?

packages/java-parser/test/samples-spec.js
```diff
@@ -9,6 +9,7 @@ const javaParser = require("../src/index");
 
 describe("The Java Parser", () => {
   createSampleSpecs("java-design-patterns");
+  createSampleSpecs("guava");
```
bd82: A failing test is a start but not too useful.

Could we modify the "createSamplesSpec" to only fail **after**
N files have failed? that could be a naive (and imperfect) way to "fix" a current state and avoid regressions.

We could even create a new createFailingSampleSpec("guava", 101)
API specifically for that scenario that will parse all the sample's file in a **single** spec.

If we can do that than we can merge this PR and incrementally work on reducing the number of failures in multiple small PRs.

bd82:
See comments in the review

bd82:
@clement26695 do you want to take a look at integrating 
these tests in a way that allows partial failures (see my comments above).

This could provide you with scenarios (bugs) to explore the parser more deeply 😄

bd82:
> Did you mean something like #134 ?

Yes.
Closing in favor of #134

PR #132, opened 2019-01-30
fix FloatLiteral pattern

### LINKED ISSUE #131: Can't parse scientific notation: 1.0e-10
I believe this is trying to parse `1.0e-2`.

Seen in Guava:

https://github.com/google/guava/blob/838560034dfaa1afdf51a126afe6b8b8e6cce3dd/guava-tests/test/com/google/common/math/DoubleMathTest.java#L715

clement26695:
The FloatLiteral pattern could now parse scientific notation as 1.0e-10 or 1.e-1

Fix #131

### DIFF
--- a/packages/java-parser/src/tokens.js
+++ b/packages/java-parser/src/tokens.js
@@ -122,9 +122,8 @@ createToken({ name: "OctalLiteral", pattern: /0_*[0-7]([0-7_]*[0-7])?[lL]?/ });
 createToken({
   name: "FloatLiteral",
   pattern: MAKE_PATTERN(
-    "\\.{{Digits}}({{ExponentPart}})?({{FloatTypeSuffix}})?|" +
-      "{{Digits}}\\.({{Digits}})?({{ExponentPart}})?({{FloatTypeSuffix}})?|" +
-      "{{Digits}}\\.({{Digits}})?({{FloatTypeSuffix}})?|" +
+    "{{Digits}}\\.({{Digits}})?({{ExponentPart}})?({{FloatTypeSuffix}})?|" +
+      "\\.{{Digits}}({{ExponentPart}})?({{FloatTypeSuffix}})?|" +
       "{{Digits}}{{ExponentPart}}({{FloatTypeSuffix}})?|" +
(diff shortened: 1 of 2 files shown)


bd82:
Thanks 👍  I will have a look at this tonight.

packages/java-parser/src/tokens.js
```diff
@@ -122,8 +122,9 @@ createToken({ name: "OctalLiteral", pattern: /0_*[0-7]([0-7_]*[0-7])?[lL]?/ });
 createToken({
   name: "FloatLiteral",
   pattern: MAKE_PATTERN(
-    "{{Digits}}\\.({{Digits}})?({{FloatTypeSuffix}})?|" +
-      "\\.{{Digits}}({{ExponentPart}})?({{FloatTypeSuffix}})?|" +
+    "\\.{{Digits}}({{ExponentPart}})?({{FloatTypeSuffix}})?|" +
```
bd82: Lets try to refactor this to align strongly with the specifications.
https://docs.oracle.com/javase/specs/jls/se11/html/jls-3.html#jls-FloatingPointLiteral

Examples:
- The Spec has 4 variations but now we have 5.
- In the spec the first variant starts with a digit, but our first variant starts with a "dot".

The stronger we align with the spec's **style** the easier it is to prove
that we are aligned with the spec's actual meaning...

clement26695:
Yes, you're right : I think it should be ok now !

bd82:
Looks good 👍

bd82:
Have a look here:
https://github.com/bd82/toml-tools/blob/985568629950665054ac56fd2601d465cb9aa716/packages/lexer/lib/tokens.js#L176-L186

I think I've found a better way to compose regExps using [Tagged Template Literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_templates).

Maybe you would like to refactor tokens.js here to use this pattern
which is more readable and allows better IDE integration as it is not all pure strings...

bd82:
Please resolve the merge conflict and we can than merge it 😄

PR #133, opened 2019-01-30
Fix parsing trailing commas in enum bodies

### LINKED ISSUE #130: Can't parse comma-semicolon in enum bodies: enum { A,; void b() {}}
I believe this is failing to parse:
```java
enum Foo {
 BAR,
;
 void baz() {}
}
```

Failure seen here, in Guava:

https://github.com/google/guava/blob/838560034dfaa1afdf51a126afe6b8b8e6cce3dd/guava-tests/test/com/google/common/primitives/ImmutableLongArrayTest.java#L247-L252

clement26695:
I tried to understand how the parser works and to fix #130. I would be happy to update the PR if something is wrong/missing !

### DIFF
--- a/packages/java-parser/src/productions/classes.js
+++ b/packages/java-parser/src/productions/classes.js
@@ -606,9 +606,14 @@ function defineRules($, t) {
   // https://docs.oracle.com/javase/specs/jls/se11/html/jls-8.html#jls-EnumConstantList
   $.RULE("enumConstantList", () => {
     $.SUBRULE($.enumConstant);
-    $.MANY(() => {
-      $.CONSUME(t.Comma);
-      $.SUBRULE2($.enumConstant);
+    $.MANY({
+      GATE: () =>
+        $.LA(1).tokenType === t.Comma &&
+        ($.LA(2).tokenType === t.Identifier || $.LA(2).tokenType === t.At),
+      DEF: () => {
+        $.CONSUME(t.Comma);
+        $.SUBRULE2($.enumConstant);
+      }
     });
   });
 
--- a/packages/java-parser/test/bugs-spec.js
+++ b/packages/java-parser/test/bugs-spec.js
@@ -9,3 +9,10 @@ describe("The Java Parser fixed bugs", () => {
     expect(() => javaParser.parse(input, "expression")).to.not.throw();
   });
 });
+
+describe("The Java Parser fixed bugs", () => {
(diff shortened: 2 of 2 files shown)


packages/java-parser/src/productions/classes.js
```diff
@@ -606,9 +606,14 @@ function defineRules($, t) {
   // https://docs.oracle.com/javase/specs/jls/se11/html/jls-8.html#jls-EnumConstantList
   $.RULE("enumConstantList", () => {
     $.SUBRULE($.enumConstant);
-    $.MANY(() => {
-      $.CONSUME(t.Comma);
-      $.SUBRULE2($.enumConstant);
+    $.MANY({
+      GATE: () =>
+        $.LA(1).tokenType === t.Comma &&
```
bd82: The Gate is an **additional** constraint
So the $.LA(1).tokenType === t.Comma is already checked.

It only has a very small runtime const, but maybe it is best to avoid unneeded checks

packages/java-parser/src/productions/classes.js
```diff
@@ -606,9 +606,14 @@ function defineRules($, t) {
...
+        ($.LA(2).tokenType === t.Identifier || $.LA(2).tokenType === t.At),
```
bd82: comparing a tokenType directly to an Identifier is unsafe and would miss some edge cases.
That is because Identifier is used as a Token Category.
- https://sap.github.io/chevrotain/docs/features/token_categories.html
- Fix: use tokenMatcher: https://github.com/jhipster/prettier-java/blob/1a4f6525ce2b86be36459e7e8db5252d94932d39/packages/java-parser/src/productions/interfaces.js#L372
- https://github.com/jhipster/prettier-java/blob/1a4f6525ce2b86be36459e7e8db5252d94932d39/packages/java-parser/src/tokens.js#L177-L179

bd82:
Also note just for reference that we could use reflection on the grammar structure to "understand"
That the first TokenType of enumConstant is either "@" or "Identifier"

That is not needed here, but good to know:
See example:
https://github.com/jhipster/prettier-java/blob/master/packages/java-parser/src/productions/expressions.js#L591-L601

clement26695:
Nice tip ! Thank you !

FauxFaux:
Nice work! I actually was planning to fix these, but haven't had any work time in.. nine days? Argh! :scream:

PR #134, opened 2019-02-01
feat(tests): create createFailingSampleSpec

clement26695:
Create function createFailingSampleSpec to only fail after N files fail (it parse all sample files in one single spec)

Related to #128

### DIFF
--- a/packages/java-parser/test/samples-spec.js
+++ b/packages/java-parser/test/samples-spec.js
@@ -10,6 +10,7 @@ const javaParser = require("../src/index");
 describe("The Java Parser", () => {
   createSampleSpecs("java-design-patterns");
   createSampleSpecs("spring-boot");
+  createFailingSampleSpec("guava", 36, 0);
 });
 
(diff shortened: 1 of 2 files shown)


packages/java-parser/test/samples-spec.js
```diff
@@ -32,3 +33,39 @@ function createSampleSpecs(sampleName) {
...
+        const sampleText = fs.readFileSync(fileDesc.path, "utf8");
+
+        try {
+          javaParser.parse(sampleText);
+        } catch (error) {
+          errors.push(relativePath);
+        }
+      });
+
+      expect(errors.length, `Can not Parse ${errors}`).to.be.at.most(
```
bd82: lets make it **exactly** the number of expected errors, so each time we fix a bug
we would have to adjust the number of failures downwards.

bd82:
I removed testing on nodejs 6.00
I do not know nor care why it failed on that version as it will soon reach EOL.

bd82:
👍

PR #135, opened 2019-02-01
fix(parser): fixmissing semicolon consuming in break statement

clement26695:
I started working on test failures on Guava samples. The first I tried to tackle is https://github.com/google/guava/blob/838560034dfaa1afdf51a126afe6b8b8e6cce3dd/android/guava/src/com/google/common/cache/Striped64.java#L216

It seems for me that there were an issue in breakStatement : the semiColon was not consumed. 
(https://docs.oracle.com/javase/specs/jls/se11/html/jls-14.html#jls-BreakStatement)


bd82:
Good catch and easy to verify because that rule is now perfectly aligned to the spec.
👍

PR #137, opened 2019-02-04
Fix dim exprs rule

clement26695:
Use backtracking in dimExprs rule because dims were confused as dimExpr in the parser (https://docs.oracle.com/javase/specs/jls/se11/html/jls-15.html#jls-15.10.1)

Fix parsing of https://github.com/google/guava/blob/838560034dfaa1afdf51a126afe6b8b8e6cce3dd/guava-tests/benchmark/com/google/common/math/StatsBenchmark.java#L149

Include #136 commits to avoid merge conflict in packages/java-parser/test/samples-spec.js

### DIFF
--- a/packages/java-parser/src/productions/expressions.js
+++ b/packages/java-parser/src/productions/expressions.js
@@ -443,6 +443,7 @@ function defineRules($, t) {
     });
   });
 
+  // https://docs.oracle.com/javase/specs/jls/se11/html/jls-15.html#jls-15.10.1
   $.RULE("arrayCreationExpression", () => {
     $.CONSUME(t.New);
     $.OR([
@@ -462,6 +463,7 @@ function defineRules($, t) {
     ]);
   });
 
+  // https://docs.oracle.com/javase/specs/jls/se11/html/jls-15.html#jls-15.10.1
   $.RULE("arrayCreationDefaultInitSuffix", () => {
     $.SUBRULE($.dimExprs);
     $.OPTION(() => {
@@ -469,18 +471,22 @@ function defineRules($, t) {
     });
   });
 
+  // https://docs.oracle.com/javase/specs/jls/se11/html/jls-15.html#jls-15.10.1
   $.RULE("arrayCreationExplicitInitSuffix", () => {
     $.SUBRULE($.dims);
     $.SUBRULE($.arrayInitializer);
   });
 
+  // https://docs.oracle.com/javase/specs/jls/se11/html/jls-15.html#jls-15.10.1
   $.RULE("dimExprs", () => {
     $.SUBRULE($.dimExpr);
-    $.OPTION(() => {
-      $.SUBRULE2($.dimExpr);
+    $.OPTION({
+      GATE: $.BACKTRACK($.dimExpr),
+      DEF: () => $.SUBRULE2($.dimExpr)
     });
   });
 
+  // https://docs.oracle.com/javase/specs/jls/se11/html/jls-15.html#jls-15.10.1
   $.RULE("dimExpr", () => {
     $.MANY(() => {
       $.SUBRULE($.annotation);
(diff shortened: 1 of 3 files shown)


bd82:
I will try to have a look later today.
Note that the expressions do not align strongly with the spec's style.
Also note that backtracking in the expressions (hot spot) could have performance implications
So if a different kind of GATE could be used that would be preferred.

clement26695:
Thank you for your comments:

> Also note that backtracking in the expressions (hot spot) could have performance implications

I know that backtracking could have performance issue but I didn't see another way to deal with the `MANY` operation at the beginnig of the dimExprs rule...

> Note that the expressions do not align strongly with the spec's style.

Why do expressions not align with the spec ? Was it done on purpose or should it be corrected ?

bd82:
I've started documenting some of the design choices here:
https://github.com/jhipster/prettier-java/blob/master/packages/java-parser/docs/design.md

Hopefully I will have time tomorrow to complete it.

> I know that backtracking could have performance issue but I didn't see another way to deal with the MANY operation at the beginnig of the dimExprs rule...

I did not look into this specific case yet, sometimes we can do optimized backtracking
by looking ahead only until a distinguishing token is found.
For example: https://github.com/jhipster/prettier-java/blob/master/packages/java-parser/src/productions/classes.js#L654-L683

You can evaluate the need for such optimize backtracking by inspecting the performance impact 
caused by adding the un-optimized backtracking, for example by using this benchmark:
https://github.com/jhipster/prettier-java/blob/master/packages/java-parser/scripts/parse-samples.js

> Why do expressions not align with the spec ? Was it done on purpose or should it be corrected ?

There are too many conflicts there to have a good alignment with adequate performance.
I will add a section on this in the design docs...

bd82:
## Back to the original PR topic

So I've taken a quick peek and it looks like to me that Dims and DimExp could be distinguished more easily.

because one has an expression inside and the other is just two tokens ("[" "]").
So perhaps instead of using backtracking when can ask a question like:
> $.LA(2).tokenType !== t.RSquare ?

clement26695:
I made the change : I thought about it, but I was confused with these lines (491->493)

```
$.MANY(() => {
  $.SUBRULE($.annotation);
});
```
But I guess annotation can not be present after the first dimExpr

packages/java-parser/src/productions/expressions.js
```diff
@@ -462,25 +463,30 @@ function defineRules($, t) {
     ]);
   });
 
+  // https://docs.oracle.com/javase/specs/jls/se11/html/jls-15.html#jls-15.10.1
```
bd82: I personally find these links to the spec a little confusing
because this does not align to the spec style strongly.

But I probably won't be maintaining this code in the long term so as this is subjective
You should decide if to keep the links, remove them or add a disclaimer to them. 😄

packages/java-parser/test/samples-spec.js
```diff
@@ -10,7 +10,7 @@ const javaParser = require("../src/index");
 describe("The Java Parser", () => {
   createSampleSpecs("java-design-patterns");
   createSampleSpecs("spring-boot");
-  createFailingSampleSpec("guava", 22, 0);
+  createFailingSampleSpec("guava", 10, 0);
```
bd82: good, almost there for the Guava code 👍 

All our three samples together are now close to 700,000 Lines of code (excluding comments and blank lines).
So I think the next steps would be to start actually implementing he prettier printer.

There is a full E2E example in Toml Prettier Plugin
- https://github.com/bd82/toml-tools/tree/master/packages/prettier-plugin-toml

And place-holders in this repo's prettier-plugin
With a micro debug able example:
https://github.com/jhipster/prettier-java/tree/master/packages/prettier-plugin-java/test/hello-world

Issue #138, opened 2019-02-05
Can't parse cast in: WildcardType toWildcardType = (WildcardType) to;


clement26695:
Found in https://github.com/google/guava/blob/838560034dfaa1afdf51a126afe6b8b8e6cce3dd/guava/src/com/google/common/reflect/TypeResolver.java#L141

It seems that the parser can not parse the casting operation here.
```java
WildcardType toWildcardType = (WildcardType) to;
```

The same situation seems to appear in : https://github.com/google/guava/blob/838560034dfaa1afdf51a126afe6b8b8e6cce3dd/guava/src/com/google/common/io/CharStreams.java#L76

bd82:
Also note we are using "optimized manual backtracking" for identifying a cast expression.
- https://github.com/jhipster/prettier-java/blob/master/packages/java-parser/src/productions/expressions.js#L200-L203

So the problem is probably in that logic

clement26695:
I understood why the parser failed : it is because of the use of the restricted keyword **to**. 

Just adding in the **isReferenceTypeCastExpression** rule a test to return true if the next token is a restricted keyword could work but there is probably something that need to be fixed elsewhere (in the unaryExpressionNotPlusMinus maybe ?)

bd82:
RestrictedKeyword can already be used as an Identifier (Token Categories).
- https://github.com/jhipster/prettier-java/blob/master/packages/java-parser/src/tokens.js#L179

But the check in the optimized backtracking lookahead
- https://github.com/jhipster/prettier-java/blob/88caf362d34d7e02ad43b83a8448ac696cb7fcdd/packages/java-parser/src/productions/expressions.js#L621

Does not use the **tokenMatcher** utility and thus does not respect the token categories.

I think we need to adjust **all** token matching to use **tokenMatcher** and **never** compare the tokenType directly using "==="

bd82:
Note the restricted keywords have some special behavior that is not yet implemented.
- https://github.com/jhipster/prettier-java/blob/88caf362d34d7e02ad43b83a8448ac696cb7fcdd/packages/java-parser/src/tokens.js#L156-L157

So some restrictedKeywords are dependent on **Parser Context**.
I think this can be solved by creating many kinds of "token categories"
that could match different sets of allowed identifiers.

PR #139, opened 2019-02-06
WIP fix(parser): fix VariableArityParameter parsing

clement26695:
Fix parsing of `@Nullable Object @Nullable... errorMessageArgs` in https://github.com/google/guava/blob/838560034dfaa1afdf51a126afe6b8b8e6cce3dd/guava/src/com/google/common/base/Verify.java#L122

The parser try to parse the second @Nullable... as dims, which is incorrect

Should be the last bug in the parser for the guava repo with #137 and #138.

WIP : There is probably a better way than using backtracking here.

### DIFF
--- a/packages/java-parser/src/productions/classes.js
+++ b/packages/java-parser/src/productions/classes.js
@@ -245,8 +245,9 @@ function defineRules($, t) {
       {
         ALT: () => {
           $.SUBRULE($.unannClassOrInterfaceType);
-          $.OPTION(() => {
-            $.SUBRULE2($.dims);
+          $.OPTION({
+            GATE: $.BACKTRACK($.dims),
+            DEF: () => $.SUBRULE2($.dims)
           });
         }
       }


bd82:
I think the same ambiguity would occur in the case of primitive Type.
```javascript
        ALT: () => {
          $.SUBRULE($.unannPrimitiveType);
          $.SUBRULE($.dims);
        }
``` 

But this is more problematic as there is no optional production wrapping the dims
to introduce a GATE. but we could probably put a GATE on the whole alternative.

These kinds of ambiguities really remind me why I dislike Java 👿 
there is so much complexity and verbosity even in the grammar 😄

You should inspect the performance impact of this change to evaluate if you need to explore optimizations. One possible optimization would be to [pass an argument](https://sap.github.io/chevrotain/docs/features/parameterized_rules.html) to the rule
so we would know if we are in a context that requires the additional backtracking
check and avoid the backtracking in other contexts.

- This could be a general optimization for similar cases.

PR #140, opened 2019-02-06
Fix block and statements

clement26695:
Fix missing semiColon in throwStatement (https://docs.oracle.com/javase/specs/jls/se11/html/jls-14.html#jls-ThrowStatement) continueStatement (https://docs.oracle.com/javase/specs/jls/se11/html/jls-14.html#jls-ContinueStatement) and returnStatement (https://docs.oracle.com/javase/specs/jls/se11/html/jls-14.html#jls-ReturnStatement) specs


bd82:
Semicolons are ugly 😢 but o.k. 😄

PR #141, opened 2019-02-11
WIP: Cst printer

clement26695:
WIP on classes and types-values-and-variables

### DIFF
--- a/packages/prettier-plugin-java/src/printers/classes.js
+++ b/packages/prettier-plugin-java/src/printers/classes.js
@@ -101,10 +101,13 @@ class ClassesPrettierVisitor {
     const unannType = this.visit(ctx.unannType);
     const variableDeclaratorList = this.visit(ctx.variableDeclaratorList);
 
-    return rejectAndJoin(" ", [
-      fieldModifiers,
+    return join("", [
+      join(" ", fieldModifiers),
+      " ",
       unannType,
-      variableDeclaratorList
+      " ",
+      variableDeclaratorList,
+      ";"
     ]);
   }
 
@@ -140,23 +143,18 @@ class ClassesPrettierVisitor {
   }
 
   variableInitializer(ctx) {
-    const variableInit = ctx.expression ? ctx.expression : ctx.arrayInitializer;
-
-    return this.visit(variableInit);
+    return this.visitSingle(ctx);
   }
 
   unannType(ctx) {
-    const type = ctx.unannPrimitiveType
-      ? ctx.unannPrimitiveType
-      : ctx.unannReferenceType;
-
-    return this.visit(type);
+    return this.visitSingle(ctx);
   }
 
   unannPrimitiveType(ctx) {
-    const type = ctx.numericType ? ctx.numericType : ctx.boolean;
-
-    return this.visit(type);
+    if (ctx.numericType) {
+      return this.visitSingle(ctx);
+    }
+    return this.getSingle(ctx).image;
   }
 
   unannReferenceType(ctx) {
@@ -174,18 +172,15 @@ class ClassesPrettierVisitor {
   }
 
   unannClassType(ctx) {
-    // const identifier = ctx.Identifier[0].image;
-    // const typeArguments = this.visit(ctx.typeArguments);
-
     return "unannClassType";
   }
 
   unannInterfaceType(ctx) {
-    return "unannInterfaceType";
+    return this.visit(ctx.unannClassType);
   }
 
   unannTypeVariable(ctx) {
-    return "unannTypeVariable";
+    return this.getSingle(ctx).image;
   }
 
   methodDeclaration(ctx) {
@@ -226,7 +221,7 @@ class ClassesPrettierVisitor {
     if (ctx.unannType) {
       return this.visit(ctx.unannType);
     }
-    // public | protected | private | Synchronized | ...
+    // void
     return this.getSingle(ctx).image;
   }
 
@@ -255,20 +250,21 @@ class ClassesPrettierVisitor {
       ? concat([ctx.Identifier[0].image, "."])
       : "";
 
-    return rejectAndJoin("", [annotations, unannType, identifier, "this"]);
+    return rejectAndJoin("", [
+      join(" ", annotations),
+      unannType,
+      identifier,
+      "this"
+    ]);
   }
 
   formalParameterList(ctx) {
     const formalParameter = this.mapVisit(ctx.formalParameter);
-    return join(",", formalParameter);
+    return join(", ", formalParameter);
   }
 
   formalParameter(ctx) {
-    if (ctx.variableParaRegularParameter) {
-      return this.visit(ctx.variableParaRegularParameter);
-    }
-
-    return this.visit(ctx.variableArityParameter);
+    return this.visitSingle(ctx);
   }
 
   variableParaRegularParameter(ctx) {
@@ -277,7 +273,7 @@ class ClassesPrettierVisitor {
     const variableDeclaratorId = this.visit(ctx.variableDeclaratorId);
 
     return rejectAndJoin(" ", [
-      variableModifier,
+      join(" ", variableModifier),
       unannType,
       variableDeclaratorId
     ]);
@@ -290,9 +286,9 @@ class ClassesPrettierVisitor {
     const identifier = ctx.Identifier[0].image;
 
     return rejectAndJoin("", [
-      variableModifier,
+      join(" ", variableModifier),
       unannType,
-      annotation,
+      join(" ", annotation),
       "...",
       identifier
     ]);
@@ -316,10 +312,7 @@ class ClassesPrettierVisitor {
   }
 
   exceptionType(ctx) {
-    if (ctx.classType) {
-      return this.visit(ctx.classType);
-    }
-    return this.visit(ctx.typeClass);
+    return this.visitSingle(ctx);
   }
 
   methodBody(ctx) {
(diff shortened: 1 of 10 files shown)


packages/prettier-plugin-java/src/printers/classes.js
```diff
@@ -50,23 +50,32 @@ class ClassesPrettierVisitor {
   }
 
   typeParameters(ctx) {
-    return "TBD";
+    const typeParameterList = this.visit(ctx.typeParameterList);
```
bd82: Maybe call it "optionalTypeParameterList" so it is easier to understand the use of "reject"?

packages/prettier-plugin-java/src/printers/classes.js
```diff
@@ -50,23 +50,32 @@ class ClassesPrettierVisitor {
...
+
+    return rejectAndJoin("", ["<", typeParameterList, ">"]);
```
bd82: maybe we should define a separate "compact" function?
- https://lodash.com/docs/4.17.11#compact

bd82: maybe we should define our own wrappers on the DOC builders of prettier?
- always compact?
- always check for invalid arguments and throw errors?
  - I often had a bug poping up at  much later stage due to passing some null  value to the doc utils
     so it was hard to debug.

packages/prettier-plugin-java/src/printers/classes.js
```diff
@@ -88,63 +97,98 @@ class ClassesPrettierVisitor {
   }
 
   fieldDeclaration(ctx) {
-    return "fieldDeclaration";
+    const fieldModifiers = this.mapVisit(ctx.fieldModifier);
+    const unannType = this.visit(ctx.unannType);
+    const variableDeclaratorList = this.visit(ctx.variableDeclaratorList);
+
+    return join("", [
```
bd82: - why join with empty string?
  - just concat instead in those situations?
- why explicitly specify whitespace?

packages/prettier-plugin-java/src/printers/classes.js
```diff
@@ -88,63 +97,98 @@ class ClassesPrettierVisitor {
...
+      return this.visit(ctx.annotation);
+    }
+    // public | protected | private | ...
+    return this.getSingle(ctx).image;
   }
 
   variableDeclaratorList(ctx) {
-    return "variableDeclaratorList";
+    const variableDeclarators = this.mapVisit(ctx.variableDeclarator);
+    const variableDeclaratorsSep = variableDeclarators.length > 0 ? ", " : "";
```
bd82: why check length if according to the grammar there is always at least one?

bd82: also why do we care about the separator? won't join on empty arraay return an empty Doc?

packages/prettier-plugin-java/src/printers/classes.js
```diff
@@ -88,63 +97,98 @@ class ClassesPrettierVisitor {
...
 
   unannReferenceType(ctx) {
-    return "unannReferenceType";
+    const type = ctx.unannPrimitiveType
+      ? this.visit(ctx.unannPrimitiveType)
+      : this.visit(ctx.unannClassOrInterfaceType);
+
+    const dims = this.visit(ctx.dims);
+
+    return rejectAndJoin("", [type, dims]);
```
bd82: we want the compact, not the join here, so lets create a compact utility?

packages/prettier-plugin-java/src/printers/classes.js
```diff
@@ -174,51 +218,109 @@ class ClassesPrettierVisitor {
...
 
   methodDeclarator(ctx) {
-    return "methodDeclarator";
+    const identifier = ctx.Identifier[0].image;
+    const receiverParameter = this.visit(ctx.receiverParameter);
+    const separator = receiverParameter ? ", " : "";
+    const formalParameterList = this.visit(ctx.formalParameterList);
+    const dims = this.visit(ctx.dims);
+
+    return rejectAndJoin("", [
```
bd82: need a  pure compact here

packages/prettier-plugin-java/src/printers/classes.js
```diff
@@ -174,51 +218,109 @@ class ClassesPrettierVisitor {
...
 
   receiverParameter(ctx) {
-    return "receiverParameter";
+    const annotations = this.mapVisit(ctx.annotation);
+    const unannType = this.visit(ctx.unannType);
+    const identifier = ctx.Identifier
+      ? concat([ctx.Identifier[0].image, "."])
+      : "";
+
+    return rejectAndJoin("", [
```
bd82: compact

packages/prettier-plugin-java/src/printers/types-values-and-variables.js
```diff
@@ -1,81 +1,194 @@
...
+            currentSegmentIdentifier,
+            currentSegmentTypeArgument
+          )
+        );
+      }
+
+      return join(".", segments);
+    }
+
+    const annotations = this.mapVisit(ctx.annotation);
```
bd82: why is the case of less complex name (no dots) not handled as part of the complex case?

packages/prettier-plugin-java/src/printers/types-values-and-variables.js
```diff
@@ -1,81 +1,194 @@
...
+      const annotations = ctx.annotation;
+      const typeArguments = ctx.typeArguments;
+      const identifiers = ctx.Identifier;
+
+      const segments = [];
+      for (const dot in dots) {
+        const offset = dot.startOffset;
+
+        const currentSegmentAnnotations = [];
+        while (annotations.length > 0) {
```
bd82: i think "annotations" could be undefined as it is optional

packages/prettier-plugin-java/src/printers/types-values-and-variables.js
```diff
@@ -1,81 +1,194 @@
...
   classOrInterfaceType(ctx) {
-    return "classOrInterfaceType";
+    return this.visitSingle(ctx);
   }
 
   classType(ctx) {
-    return "classType";
+    const dots = ctx.Dots;
+
+    if (dots) {
```
bd82: maybe if you sort **everything** (except dots) by the offset first you would have an easy to work with structure?
it would look very much like the grammar.
> (annotation)* identifier (typeArguments)?

packages/prettier-plugin-java/src/printers/types-values-and-variables.js
```diff
@@ -1,81 +1,194 @@
...
   dims(ctx) {
+    // TODO
     return "dims";
   }
 
   typeParameter(ctx) {
-    return "typeParameter";
+    const typeParameterModifiers = this.mapVisit(ctx.typeParameterModifier);
+    const typeParameterModifiersOutput =
+      typeParameterModifiers.length > 0
```
bd82: either its undefined or with length > 0
if its undefined there would be a runtime error that .length property does not exist on undefined.

optional productions are undefined in the CST

packages/prettier-plugin-java/src/printers/types-values-and-variables.js
```diff
@@ -1,81 +1,194 @@
...
-    return "typeArguments";
+    const typeArgumentList = this.visit(ctx.typeArgumentList);
+
+    return concat(["<", typeArgumentList, ">"]);
   }
 
   typeArgumentList(ctx) {
-    return "typeArgumentList";
+    const typeArguments = this.mapVisit(ctx.typeArgument);
+    const typeArgumentSep = typeArguments.length > 0 ? ", " : "";
```
bd82: there is never length === 0

packages/prettier-plugin-java/test/hello-world/_input.java
```diff
@@ -1,5 +1,26 @@
 package graphics. x.y   .z ;
 
-public   class    HelloWorld {
+public   class    HelloWorld <AR     >    extends
```
bd82: ouch this is badly formatted 😄

bd82:
focus on defining the "best" patterns and utils now before implementing too much code that it would be harder to refactor later 😄

clement26695:
I have added a rejectAndConcat util. I was wondering : should we always use the rejectAndJoin and rejectAndConcat utils instead of join and concat ? Or should we use it only when there could be a falsey value ?

packages/prettier-plugin-java/src/printers/types-values-and-variables.js
clement26695: I thought about it. What could be difficult is that we can only have information on tokens' offset : it would be more difficult to know the offset of a rule, unless we start exploring its children...

bd82:
> I have added a rejectAndConcat util. I was wondering : should we always use the rejectAndJoin and rejectAndConcat utils instead of join and concat ? Or should we use it only when there could be a falsey value ?

I think that perhaps we could make our own concat and join that always implicitly reject as it seems to be a reoccurring pattern.

This means never using the prettier docBuilders directly and instead using
our "more idiot proof" builder wrapper.

Also @clement26695 I think we should start merging this work so that new commits 
would each add a small "printer" implementation...
Otherwise this would grow to a 4,000 lines commits 😢  

@jdubois Can you give @clement26695 write permissions to this repository so he can push directly to master?

bd82:
Can I merge this @clement26695

packages/prettier-plugin-java/src/printers/types-values-and-variables.js
bd82: Yes but exploring the first element of annotation or typeArguments to determine their offset should be trivial, we do not need a generic solution for offset range for any and all CSTNodes.

clement26695:
It is ok for me to merge this. I may have to refactor some code regarding your comments, but it could be done later.

I'm also ok for systematically using our own implementation of join and concat

PR #142, opened 2019-03-06
Fix annotation: accepting comma as last element in ElementValueList

Shaolans:
As I was trying to parse the spring-framework repository, this file was failing to parse https://github.com/spring-projects/spring-framework/blob/master/spring-context/src/test/java/org/springframework/context/annotation/PropertySourceAnnotationTests.java#L442 (L442 and L451) due to the metadata content in the annotation. It seems that JAVA accept ElementValueList to end with a comma even though the JAVA specification says otherwise https://docs.oracle.com/javase/specs/jls/se11/html/jls-9.html#jls-ElementValueList.

### DIFF
--- a/packages/java-parser/src/productions/interfaces.js
+++ b/packages/java-parser/src/productions/interfaces.js
@@ -305,7 +305,13 @@ function defineRules($, t) {
     $.SUBRULE($.elementValue);
     $.MANY(() => {
       $.CONSUME(t.Comma);
-      $.SUBRULE2($.elementValue);
+      /*
+      To be compliant with the JAVA specification we should remove the Option.
+      However, JAVA accept that ElementValueList ends with a comma (,)
+      */
+      $.OPTION(() => {
+        $.SUBRULE2($.elementValue);
+      });
     });
   });
 
--- a/packages/java-parser/src/productions/interfaces.js
+++ b/packages/java-parser/src/productions/interfaces.js
@@ -303,15 +303,12 @@ function defineRules($, t) {
   // https://docs.oracle.com/javase/specs/jls/se11/html/jls-9.html#jls-ElementValueList
   $.RULE("elementValueList", () => {
     $.SUBRULE($.elementValue);
-    $.MANY(() => {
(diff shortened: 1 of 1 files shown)


bd82:
The dangling comma already appears in the **ElementValueArrayInitializer** rule.

https://github.com/jhipster/prettier-java/pull/142/files#diff-995f4f77eccdd18db1ee0036c282e385R298

But there is an ambiguity here as the parser thinks the next comma is the start of another value
instead of identifying it as a ***dangling comma***

There are a few ways to resolve this:
1. https://github.com/bd82/toml-tools/blob/508d5d5d13e8cf765b7c87a45f569f4859478995/packages/parser/lib/parser.js#L98-L116
1. In the **ElementValueList** repetition, add a GATE that inspects 2 tokens ahead ($.LA(2).tokenType === ...)

Shaolans:
Thanks ! I have now a better understanding of the issue, I tried to resolve this with a GATE as you suggested and it seems to parse the file now.
I am still not sure if this fix is correct or is what you meant to do, I am still pretty new with Chevrotain :)

bd82:
> I am still not sure if this fix is correct or is what you meant to do, I am still pretty new with Chevrotain :)

The fix is perfect.

If the dangling comma was specified in the same rule than Chevrotain would have automatically looked two tokens ahead to distinguish between the alternatives.

However because Chevrotain is ***[context free](https://en.wikipedia.org/wiki/Context-free_grammar)*** parsing library the rule stack (part of the context) is not taken into account  when performing the decision to go into the the repetition.

And because we are trying to align strongly with the ***style of the official specs*** we can't move the dangling comma definition around much.

Chevrotain Parsers are normally a-lot simpler.
- e.g [ECMAScript 5 grammar](https://github.com/SAP/chevrotain/blob/master/examples/grammars/ecma5/ecma5_parser.js)

This project contains the most complex Chevrotain parser I am aware of.
That is mostly due to the complexity of the Java Grammar and partially due to the constraint of aligning with the spec style. 😢

PR #144, opened 2019-03-07
Few removal of TODOs and small upgrades

Hawkurane:
Hello, I've been reviewing the code and and seen a few TODOs that were easily doable (unless I've misunderstood them). Also added a few repos to the sampling to make sure the parsing works as intended with the applications JHipster can generate.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 /packages/java-parser/samples
+node_modules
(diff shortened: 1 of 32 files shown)


Hawkurane:
Hello again, according to Shaolans, I've integrated a bad fix ( #142 ) into my pull request. I am currently removing the commit from my work and will once open a pull request. Also, would you like me to open a pull request for every todo?

bd82:
One PR for every tiny change may be too much.
You can still have multiple small commits in a PR, but just squash them together during merge (there is a button for that on the PR). so the history of the repository would not get polluted with too much junk.

On the other hand you obviously want to avoid too large PRs so keep that in mind too 😄

Issue #105, opened 2018-11-13
Align implementation of whitespace Token with the Spec


bd82:
https://github.com/jhipster/prettier-java/blob/5ef66ae309b41e6634d6e027631c824482d4af96/packages/java-parser/src/tokens.js#L108-L109

"\s" regexp in JavaScript has a very specific meaning.
>  Matches a single white space character, including space, tab, form feed, line feed and other Unicode spaces. Equivalent to [ \f\n\r\t\v\u00a0\u1680\u2000-\u200a\u2028\u2029\u202f\u205f\u3000\ufeff].      For example, /\s\w*/ matches " bar" in "foo bar".

But that may not be equivalent to the definition of **Java** whitespace.

bd82:
See the offical definition in the Java Specifications.
https://docs.oracle.com/javase/specs/jls/se11/html/jls-3.html#jls-3.

Hawkurane:
Hello, I've submitted a pull request (#145) to redefine the whitespace definition after adding a few more sample repos to parse (mainly sample JHipster apps and spring framework to make sure it can run at least through them).

bd82:
Hi @Hawkurane and thanks 😄 

I reviewed #145 See comments there

clement26695:
Close this as it is resolved in #146 (correct me if I'm wrong)

PR #145, opened 2019-03-07
Whitespace redefinition

Hawkurane:
Hello, creating a pull request to add more sampling data (repos mainly from jhipster sample applications and spring framework) and reviewing the whitespace definition to fit the Java Spec's one as referenced in issue https://github.com/jhipster/prettier-java/issues/105

### DIFF
--- a/packages/java-parser/src/tokens.js
+++ b/packages/java-parser/src/tokens.js
@@ -140,7 +140,7 @@ createToken({
 });
 createToken({
   name: "DecimalLiteral",
-  pattern: MAKE_PATTERN("(0|[1-9](({{Digits}})?|_+{{Digits}}))[lL]?")
+  pattern: MAKE_PATTERN("(0|[1-9](_+{{Digits}}|({{Digits}})?))[lL]?")
 });
 // https://docs.oracle.com/javase/specs/jls/se11/html/jls-3.html#jls-3.10.4
 createToken({


packages/java-parser/scripts/single-sample-runner.js
```diff
@@ -11,3 +11,4 @@ public @interface ConfigurationProperties {}
 `;
 
 javaParserChev.parse(input, "typeDeclaration");
+console.log();
```
bd82: this should not be here

packages/java-parser/scripts/parse-samples.js
```diff
@@ -22,53 +22,105 @@ if (options.printErrors) {
   printErrors = console.error;
 }
 
-// const samplesDir = path.resolve(__dirname, "../samples/java-design-patterns");
-const samplesDir = path.resolve(__dirname, "../samples/spring-boot");
-const sampleFiles = klawSync(samplesDir, { nodir: true });
-const javaSampleFiles = sampleFiles.filter(fileDesc =>
-  fileDesc.path.endsWith(".java")
+const samplesDir = [];
```
bd82: you need to update the sample spec
- https://github.com/jhipster/prettier-java/blob/master/packages/java-parser/test/samples-spec.js

Not add every single sample here directly.
This just a dev tool, e.g for local benchmarking

packages/java-parser/src/tokens.js
```diff
@@ -30,6 +30,7 @@ FRAGMENT("ExponentPart", "[eE][+-]?{{Digits}}");
 FRAGMENT("HexDigit", "[0-9a-fA-F]");
 FRAGMENT("HexDigits", "{{HexDigit}}(({{HexDigit}}|'_')*{{HexDigit}})?");
 FRAGMENT("FloatTypeSuffix", "[fFdD]");
+FRAGMENT("LineTerminator", "(\\x0A|(\\x0D(\\x0A)?))");
```
bd82: I think fixing the whitespace should be in a separate PR from the sample updates

PR #146, opened 2019-03-08
Aligned whitespace definition with Java Specs

Hawkurane:
As suggested in (https://github.com/jhipster/prettier-java/pull/145) I've re-opened a PR with only changing the whitespace redefinition.

### DIFF
--- a/packages/java-parser/src/tokens.js
+++ b/packages/java-parser/src/tokens.js
@@ -30,6 +30,7 @@ FRAGMENT("ExponentPart", "[eE][+-]?{{Digits}}");
 FRAGMENT("HexDigit", "[0-9a-fA-F]");
 FRAGMENT("HexDigits", "{{HexDigit}}(({{HexDigit}}|'_')*{{HexDigit}})?");
 FRAGMENT("FloatTypeSuffix", "[fFdD]");
+FRAGMENT("LineTerminator", "(\\x0A|(\\x0D(\\x0A)?))");
 
 const Identifier = createTokenOrg({
   name: "Identifier",
@@ -105,8 +106,15 @@ const UnarySuffixOperator = createToken({
   pattern: Lexer.NA
 });
 
-// TODO: align with Java Spec
-createToken({ name: "WhiteSpace", pattern: /\s+/, group: Lexer.SKIPPED });
+// https://docs.oracle.com/javase/specs/jls/se11/html/jls-3.html#jls-3.6
+// Note [\\x09\\x20\\x0C] is equivalent to [\\t\\x20\\f] and that \\x20 represents
+// space character
+createToken({
+  name: "WhiteSpace",
+  pattern: MAKE_PATTERN("[\\x09\\x20\\x0C]|{{LineTerminator}}"),
+  group: Lexer.SKIPPED
+});
+
 createToken({
   name: "LineComment",
   pattern: /\/\/[^\n\r]*/,


packages/java-parser/src/tokens.js
```diff
@@ -105,8 +106,11 @@ const UnarySuffixOperator = createToken({
   pattern: Lexer.NA
 });
 
-// TODO: align with Java Spec
-createToken({ name: "WhiteSpace", pattern: /\s+/, group: Lexer.SKIPPED });
+createToken({
```
bd82: add a comment linking to: https://docs.oracle.com/javase/specs/jls/se11/html/jls-3.html#jls-3.6
above the token def.

packages/java-parser/src/tokens.js
```diff
@@ -105,8 +106,11 @@ const UnarySuffixOperator = createToken({
...
+  name: "WhiteSpace",
+  pattern: MAKE_PATTERN("[\\x09\\x20\\x0C]|{{LineTerminator}}"),
```
bd82: Do you think it would be more readable to use specific escapes where applicable?
```javascript
var hardForMeWithHexEscapes = '[\\x09\\x20\\x0C]';
// left the hex escape for space cause otherwise it can be "missed" while reading.
var easierForMe = '[\\t\\x20\\f]'

const space = " ";
var evenEasierToRead = `[\\t${space}\\f]`;
```

bd82: Also have a look here:
https://github.com/bd82/toml-tools/blob/master/packages/lexer/lib/tokens.js#L72

I've implemented a better DSL for composing regExps in Toml Tools Parser.
This should not be integrated as part of this change of course, but if you find this useful it could be
considered for integration in a separate PR.

packages/java-parser/src/tokens.js
```diff
@@ -30,6 +30,7 @@ FRAGMENT("ExponentPart", "[eE][+-]?{{Digits}}");
 FRAGMENT("HexDigit", "[0-9a-fA-F]");
 FRAGMENT("HexDigits", "{{HexDigit}}(({{HexDigit}}|'_')*{{HexDigit}})?");
 FRAGMENT("FloatTypeSuffix", "[fFdD]");
+FRAGMENT("LineTerminator", "(\\x0A|(\\x0D(\\x0A)?))");
```
bd82: See comment about hex escapes vs specific escapes below ("\n" vs "\x0A")

bd82:
We can merge this.

I will leave for you to decide if the readability related comments (hex vs specific escapes) are important enough to resolve in this PR / in a future commit / or never.

😄

packages/java-parser/src/tokens.js
Hawkurane: I think for readability, it is better to use specific escapes, but personnally, aesthetically, I don't like mixing up specific escapes and hex escapes.
I tried to do the "evenEasierToRead" method since I do quite like it, but it seems to make the Lexer crash.
Instead we could let the Hex escapes and comment the equivalent escape so anyone can read depending on their preferences.

Hawkurane:
I've added the comments in my branch.
I haven't commented the fragment because I think that the fragment name is self-explanatory.

Could you check the PR as I've also merged from upstream/master? I'm unsure if the merging in the commit history will be merged as well in the PR. If needed, I can reset hard and force push with a cleaner history.

packages/java-parser/src/tokens.js
bd82: A comment next to hex escapes is fine too.

PR #147, opened 2019-03-08
Using tokenMatcher instead of === to compare tokenType

### LINKED ISSUE #138: Can't parse cast in: WildcardType toWildcardType = (WildcardType) to;

Shaolans:
I this PR I changed the comparison of tokenType with "===" by tokenMatcher as it supports token categories and at the same time it fix #138.

### DIFF
--- a/packages/java-parser/src/productions/arrays.js
+++ b/packages/java-parser/src/productions/arrays.js
@@ -1,4 +1,7 @@
 "use strict";
+
+const { tokenMatcher } = require("chevrotain");
+
 function defineRules($, t) {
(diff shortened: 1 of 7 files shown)


bd82:
Great 👍  

This is a-lot more robust even if there may be a tiny performance hit

PR #148, opened 2019-03-08
Reinforced test sampling data and added repo's to parse

Hawkurane:
I've added sampling data for testing purposes. It mainly contains jhipster-generated applications and jhipster-related repositories to make sure it can be ran with JHipster.

### DIFF
--- a/packages/java-parser/test/samples-spec.js
+++ b/packages/java-parser/test/samples-spec.js
@@ -10,6 +10,24 @@ const javaParser = require("../src/index");
 describe("The Java Parser", () => {
   createSampleSpecs("java-design-patterns");
   createSampleSpecs("spring-boot");
+  createSampleSpecs("spring-framework");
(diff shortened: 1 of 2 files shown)


packages/java-parser/test/samples-spec.js
```diff
@@ -10,6 +10,24 @@ const javaParser = require("../src/index");
 describe("The Java Parser", () => {
   createSampleSpecs("java-design-patterns");
   createSampleSpecs("spring-boot");
+  createSampleSpecs("spring-framework");
```
bd82: **Comment maybe outside scope of this PR**

We have a **duplication** in the list of repo folder names between this and the clone samples script.
Perhaps we should extract the clone-samples repo data to a separate file
and generate these specs using that data instead of having to add a "createSampleSpec" line for each new
repo added?

That would be less error prone.

PR #152, opened 2019-03-11
Fix try with resources: acception semicolon as last token

Shaolans:
Hello, this time I tried to parse the jenkins repository and it failed at this line https://github.com/jenkinsci/jenkins/blob/master/test/src/test/java/hudson/cli/CLITest.java#L268 due to the semicolon at the end of the try with resources. The problem is similar to #142 and can be fixed with a GATE as well.


bd82:
Looks good 👍

PR #154, opened 2019-03-11
Fix brackets considered as a cast

Shaolans:
This time I was trying to parse the elasticsearch repository, I have multiple fails caused due to different bugs (8 files that fails, 5 of them are fixed with the PR #152 ). Two of them is due to this kind of line ```return (long)(left) << right;``` https://github.com/elastic/elasticsearch/blob/master/modules/lang-painless/src/main/java/org/elasticsearch/painless/DefMath.java#L964 and https://github.com/elastic/elasticsearch/blob/v6.6.1/plugins/analysis-icu/src/main/java/org/elasticsearch/index/analysis/IndexableBinaryStringTools.java#L188.
This is the rule stack:
```
[ 'typeDeclaration',
        'classDeclaration',
        'normalClassDeclaration',
        'classBody',
        'classBodyDeclaration',
        'classMemberDeclaration',
        'methodDeclaration',
        'methodBody',
        'block',
        'blockStatements',
        'blockStatement',
        'statement',
        'statementWithoutTrailingSubstatement',
        'returnStatement',
        'expression',
        'ternaryExpression',
        'binaryExpression',
        'unaryExpression',
        'primary',
        'primaryPrefix',
        'castExpression',
        'primitiveCastExpression',
        'unaryExpression',
        'primary',
        'primaryPrefix',
        'castExpression',
        'referenceTypeCastExpression',
        'unaryExpressionNotPlusMinus',
        'primary',
        'primaryPrefix',
        'fqnOrRefType',
        'fqnOrRefTypePart',
        'typeArguments',
        'typeArgumentList',
        'typeArgument' ]
```
It seems that the (long) is recognized as a cast which true but (left) is also recognized as a cast.
But this problem is only present when it is followed by ```<<``` or ```>>```.
This is due to the rule "isCastExpression" return true for (left) because **firstForUnaryExpressionNotPlusMinus** contains ">" and "<" tokens. The only fix I thought of is to manually check the next token. I don't really know if there is a better way to fix it.

### DIFF
--- a/packages/java-parser/src/productions/expressions.js
+++ b/packages/java-parser/src/productions/expressions.js
@@ -250,8 +250,10 @@ function defineRules($, t) {
     ]);
   });
 
+  // See https://github.com/jhipster/prettier-java/pull/154 to understand
+  // why fqnOrRefTypePart is split in two rules (First and Rest)
   $.RULE("fqnOrRefType", () => {
-    $.SUBRULE($.fqnOrRefTypePart);
+    $.SUBRULE($.fqnOrRefTypePartFirst);
 
     $.MANY2({
       // ".class" is a classLiteralSuffix
@@ -262,7 +264,7 @@ function defineRules($, t) {
         tokenMatcher(this.LA(2).tokenType, t.New) === false,
       DEF: () => {
         $.CONSUME(t.Dot);
-        $.SUBRULE2($.fqnOrRefTypePart);
+        $.SUBRULE2($.fqnOrRefTypePartRest);
       }
     });
 
@@ -285,7 +287,7 @@ function defineRules($, t) {
   //       3. "Super" cannot be mixed with "classTypeArguments" or "annotation".
   //       4. At most one "Super" may be used.
   //       5. "Super" may be last or one before last (last may also be first if there is only a single part).
-  $.RULE("fqnOrRefTypePart", () => {
+  $.RULE("fqnOrRefTypePartRest", () => {
     $.MANY(() => {
       $.SUBRULE($.annotation);
     });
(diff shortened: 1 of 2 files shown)


bd82:
I need to debug this to understand fully.
Also this is a none trivial edge case so it should be accompanied by a test case.
I will try to create one now.

bd82:
Here are a couple test cases you can use
```javascript

  it("cast expression, parenthesis and binary shift", () => {
    const input = "(left) << right";
    expect(() => javaParser.parse(input, "expression")).to.not.throw();
  });

  it("cast expression, parenthesis and lessThan", () => {
    const input = "(left) < right";
    expect(() => javaParser.parse(input, "expression")).to.not.throw();
  });
```

bd82:
This is quite complicated.

- Firstly I saw **firstForUnaryExpressionNotPlusMinus ** includes "<" but not ">".

The more interesting aspect is that this error is "transitive** due to an error in the original grammar.
So it is best to fix the original error.

The error is related to these lines:
https://github.com/jhipster/prettier-java/blob/83c4aaf5886fbeb2638974758003b88f1f5712e6/packages/java-parser/src/productions/expressions.js#L293-L298

Basically these "method typeArguments which originate from the [methodInovcation](https://docs.oracle.com/javase/specs/jls/se11/html/jls-15.html#jls-MethodInvocation) rule
should not be allowed for the **first** "fqnOrRefTypePart" rule.

So maybe we need to create two rules:
- "fqnOrRefTypeFirst"
- "fqnOrRefTypeRest"

Note we should not fully duplicate the rules, instead we can extract the common parts to a common
rule to avoid duplication...

Shaolans:
Hi, I think from now on, I will dig deeper instead of trying to do quick fix :s

bd82:
> Hi, I think from now on, I will dig deeper issues instead of trying to do quick fix :s

That is generally a good idea 😄 
Although issues such as this combine multiple and indirect causes which means they are fairly complex so don't be afraid to discuss them (digging too deep **alone** may be ineffective...)

Issue #116, opened 2019-01-10
Support Unicode Identifiers


bd82:
## The Problem
Java Identifiers may contain unicode letters, for example:
```java
// my name in Hebrew as the Identifier and its spelling in English as the value.
String שחר = "shahar";
```

See the relevant section in the Java Specifications:
http://m/javashttps://docs.oracle.coe/specs/jls/se11/html/jls-3.html#jls-3.8

The specs define the allowed unicode characters **indirectly** by using the results
of the Java methods:
 - Character.isJavaIdentifierStart(int)
-  Character.isJavaIdentifierPart(int)

These methods are naturally not accessible by default in a Java**Script** run time.

### Represent the unicode character rules as a regular JavaScript RegExp
This is what the Acorn & Esprima Java**Script** parsers do which creates some seriously large regExps:
- https://github.com/acornjs/acorn/blob/d38536b08e0b8a84b7e871bfc2c7dbd7a1c24d90/acorn/src/identifier.js#L24-L31
- https://github.com/jquery/esprima/blob/16aa6f168afbedb97703c896588fc05e9e56ce72/src/character.ts#L1-L8

Assuming the Java unicode rules are similar in complexity to the JavaScript unicode rules
We will be looking at creating (generating?) very large regexps...

### Implement the isJavaIdentifierXXX in JavaScript and use Custom Token Patterns.

A better solution would probably be to avoid using regExps.
Instead Chevrotain can define tokens using arbitrary JavaScript code as well.
This feature is called [Custom Token Patterns](http://sap.github.io/chevrotain/docs/features/custom_token_patterns.html)

What this means is that if we can [translate/compile/access] the Character.isJavaIdentifierXXX
in a JavaScipt runtime, we could implement the unicode Identifiers in a much cleaner fashion
without adding the abstraction of regExps which in this case seems to mostly add complexity.

bd82:
@jdubois I think this is a good candidate for the bounty program.
Should I just add the $$$ label to it?

Hawkurane:
Hello @bd82, I've managed to implement unicode identifiers through a very long and ugly regexp.
It is certainly possible to go through a Custom Token Pattern, but it would still end in matching very but shorter and still ugly regexps. What do you think we should do?

clement26695:
Close this as it is resolved in #155, #170 and #171 (correct me if I'm wrong)

PR #155, opened 2019-03-11
[WIP] Enabled identifiers with Unicode characters in lexer

Hawkurane:
Attempt to fix Issue #116 by declaring a very long regex rule that does not look like anything.


bd82:
This would be very hard to maintain. (probably impossible... 😄 )

Consider the following options or a combination of these.
- Generate this regExp from some source (if there is a machine readable definition somewhere).
- Use a custom [Token pattern](https://sap.github.io/chevrotain/docs/guide/custom_token_patterns.html) to define the Identifier pattern as a JavaScript function.
  - If it is a JavaScript function perhaps it can very closely match the Java implementation and 
     that property will provide us with some assurance of its correctness. 
- Look at Java --> JavaScript compilers, to compile those two Java Functions to JavaScript.

@jdubois Are there any **legal implications** here of re-implementing Java source code? (Oracle...)
Perhaps only OpenJDK source code should be used as a reference?

Hawkurane:
I've used a [Unicode Range Generator](https://apps.timwhitlock.info/js/regex#) to generate the regexps according to the Java Exp, it does not look like anything but it's working according to the tests.
Do you have any recommendations for Java --> JavaScript compilers ?

bd82:
>  Do you have any recommendations for Java --> JavaScript compilers ?

Not really, and those may have too much dependent code to be useful (we only need to convert a little bit of code.

What I am worried is that there is no automated way to reproduce the work you've done to create this regExp. so if the definition changes (e.g in Java 12) it would all be done from scratch and each time this manual work is performed new (human) errors may sneak in.

Hawkurane:
The main problem is to work with unicode categories (i.e. [Number Letters](https://www.compart.com/en/unicode/category/Nl)) and most of the nodes modules on npm are under the MIT License. One way would be to create our own module for use re-using the Unicode Database (ftp://ftp.unicode.org/Public/UNIDATA/UnicodeData.txt).

bd82:
Example from Esprima on dealing with complex unicode Identifiers
- https://github.com/jquery/esprima/blob/master/src/character.ts#L1-L8
- https://github.com/jquery/esprima/blob/master/tools/generate-identifier-regex.js

Does this help?

Hawkurane:
They defining authorized characters thanks to a regexp and testing one by one letters, similarly to isJavaIdentifier, but with JavaScript specs.
I already have the Regexp working and am currently working on a Custom Token solution but what I am looking for is how to go without using regexps and directly using Unicode Categories. Also, after further documentation, it seems Java does use the Unicode Database file to check if the identifiers that are using unicode characters are from a certain category or not.

bd82:
From a high level the concerns are:

- Performance
  - Should be fast for the simple case (most Identifiers are just english letters) and each Java text has a great many Identifiers.
- maintainability 
  - If its a huge regExp it should be generated somehow.
  - If it is a custom Token pattern the category ranges should be documented and readable, basically hide the magic hex values somehow.

Hawkurane:
After implementing the Custom Token solution, loading the Unicode file takes negligible time (44ms on my machine), but the parsing gets affected quite a lot. Parsing the spring-boot repository, I go from 7667ms (current master) to 13357ms (custom token). Also, using the PR's code, I get 7364ms.

I have tried to store in Sets the decimal codes of the allowed categories of Unicode Characters (checked from Java Specifications) and checked whether a character's decimal code is stored within one of them to check if it is allowed as a JavaIdentifierXXX.
Maintaining this code would probably choose which Unicode Category we would like to add up to the search.

Currently, generating the huge regexp is done with a generator that seems to be fair to the Unicode Specifications concerning categories. As long as Java specifications uses the same unicode specifications, it should be safe to use it. The error margin concerning the maintainability stays at the level where manual work is required with the help of outside tools.

bd82:
Can you share the work with the custom token pattern?
- in a separate branch for example

I think I need to look more deeply into the code to provide more productive help here.

And do you think it is possible to generate this regExp using a script in our code
by using data from the unicode metadata file?

Hawkurane:
Sure, you can check:

- [My custom pattern token (wip)](https://github.com/Hawkurane/prettier-java/blob/9a657ec327e0fd0f608b99ca787e41a977f9df30/packages/java-parser/src/tokens.js#L49)
- [Loading the Unicode Data file](https://github.com/Hawkurane/prettier-java/blob/9a657ec327e0fd0f608b99ca787e41a977f9df30/packages/java-parser/src/unicode.js#L51)

Please excuse the state of my branch, as it is where I usually play around with the code to mess with the parser

bd82:
Thanks.

I will have a look at this over the weekend

bd82:
o.k, some feedback:

## charCodeAt
[charCodeAt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charCodeAt) does not have full support for unicodeit only returns a 16bit value.
  - https://github.com/Hawkurane/prettier-java/blob/9a657ec327e0fd0f608b99ca787e41a977f9df30/packages/java-parser/src/tokens.js#L51
  You probably want to use [codePointAt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt)

## Using sets may be ineffective.
```javascript
  if (
    chars.unicode.Nl.has(charCode) ||
    chars.unicode.Lo.has(charCode) ||
    chars.unicode.Ll.has(charCode) ||
    chars.unicode.Lt.has(charCode) ||
    chars.unicode.Lm.has(charCode) ||
    chars.unicode.Lu.has(charCode) ||
    chars.unicode.Sc.has(charCode) ||
    chars.unicode.Pc.has(charCode)
  ) {
    endOffset++;
    charCode = char.charCodeAt(endOffset);
}
```

This does **a-lot** of lookups for every single character. it may be equivalent from as asymptotic
complexity perspective, but in the "real world" constant factors can make a significant difference.
- Maybe try to create only two sets, one for the first characters and one for the "rest" characters?   

## Optimize for the common case.
Most Identifiers only use a subset of ASCI characters.
You could start your custom lookahead function by first matching against `/[A-Za-Z][A-Za-z0-9]*/`
and then check the next character, if the next character is **not** a unicode identifier character
then we are done and will have almost the performance of using the regExp directly.

## Don't load the unicode categories file each time the program is executed.
- It would be difficult to bundle for browser usage.
- It would increase the size of the java-parser package unnecessarily.
- It would slow down the initialization.
- I am not sure under what license is the unicode categories file licensed, it is best if
   we can avoid 3rd party contents as part of our repository...

Instead create a script that reads the unicode categories file and generates a JavaScript file that includes that info we need.  for example:

```javascript
// This is generated code
const firstIdentChar = new Set([
97, 98, 99, 100
// ...
])

const restIDentChar = new Set(firstIdentChar.values.concat([
// more charCode here...
]))

module.exports = {
   firstIdentChar,
   restIDentChar
}
```

## Enable lexer optimizations
This is probably one of the main factors of the performance degradations.
- https://sap.github.io/chevrotain/docs/guide/performance.html#lexer-optimizations
- https://sap.github.io/chevrotain/docs/guide/custom_token_patterns.html#usage
  - See the: `start_chars_hint` property in the token configuration.

bd82:
You can find info on decoding and encoding surrogate pairs here:
https://unicodebook.readthedocs.io/unicode_encodings.html#utf-16-surrogate-pairs

Hawkurane:
Hello, I've reviewed in [my branch](https://github.com/Hawkurane/prettier-java/commit/450ff50127b2bc387c24fd8d6915f65b26a79b03) the corrections you gave to me and everything works just fine !
I am indeed able to parse spring-boot in (now) 8975ms. As I have changed everything in one go, I am unsure where we lost so much computing time in our algorithm but we are much closer than as if we were parsing using a long regexp. I have **not** tried using the lookahead function for common case yet.
I've added another repo to parse for correctness purposes, and seem to be able to parse these files:

- [File 1](https://github.com/alibaba/fastjson/blob/master/src/test/java/com/alibaba/json/bvt/bug/Issue208.java#L11)
- [File 2](https://github.com/alibaba/fastjson/blob/05fb7c5f8bb02351574f4921806092b45ce1dd90/src/test/java/com/alibaba/json/bvt/bug/Bug_for_BlankRain_Issue_502.java#L16)

Also, I've added a feature to generate the javascript file as intended, but I still do not know when to generate or not the file. At the moment, it is done when the tokens.js file is executed but I think it would make more sense to do it when we initialize the lexer?

Do you think the lookahead function to check the common case regexp would make us gain the 2 seconds lost while parsing the spring-boot repository?

bd82:
> I've added another repo to parse for correctness purposes,

Add one or two tiny unit tests for unicode identifiers (without relying on cloning a whole repo).

> Also, I've added a feature to generate the javascript file as intended, but I still do not know when to generate or not the file. At the moment, it is done when the tokens.js file is executed but I think it would make more sense to do it when we initialize the lexer?

The idea of generating the file with the metadata is to improve loading performance.
The content of this file can only change if the specification of Java Changes
or the unicode metadata changes.  Neither of these things happens often
so the generation should be done as an npm script and executed manually on demand. 

> Do you think the lookahead function to check the common case regexp would make us gain the 2 seconds lost while parsing the spring-boot repository?

There is some fixed overhead for using custom token patterns, but I think it is worth checking if we can optimize if farther. It seems easy enough to check by implementing that optimization...

> Edit: I have found the license concerning the file:

For example these terms state that you must include the license file and copyright notice in any distribution. It could be simpler to re-download the unicode categories file to a temp (ignored) folder
when running the npm script to generate the unicode Sets (in JavaScript).
That way 3rd party copyrighted code would not be part of this repository.

Hawkurane:
Ok, I've [changed the script](https://github.com/Hawkurane/prettier-java/commit/7f7a0feb4f6f01d2ca5453ef9a6f10b512b658c1) so that the scripts parses a UnicodeData.txt file given in argument. All the user needs to do is to download the file and manually specify the file to parse, everything should be fine.

As for the lookahead function I've tried a few things without any real working solutions. Do you have any documentation for me to look out for without just going with an if and a few matches in the matchIdentifierJava ? 

> Add one or two tiny unit tests for unicode identifiers (without relying on cloning a whole repo).

I wasn't planning on adding this on a future PR or modifying this PR with parsing a whole repo just to make sure I could parse chinese characters. I was just very lazy to recreate a weirdly coded Java file :joy:

However, adding this repo to the tests might be interested to make sure future changes can still parse unicode characters (Chinese characters, or more languages if you know some people that use non-latin language to code :cold_sweat:)

bd82:
>  I was just very lazy to recreate a weirdly coded Java file 😂

😄  with a unit test you won't need any java files, just strings of java source code.
e.g : https://github.com/jhipster/prettier-java/blob/master/packages/java-parser/test/bugs-spec.js#L15

> However, adding this repo to the tests might be interested to make sure future changes can still parse unicode characters

Sort of yes, but:
- We have many repos already, the amount of time it takes the tests to run is also a consideration.
- Adding a repo for every edge case is sub-optimal.
- Who guarantees that in some future update of that repo's tag the repo would still exhibit the edge case? 
  - Maybe someone will enforce ASCI identifiers if/when that project attracts multi-national contributors.

bd82:
> All the user needs to do is to download the file and manually specify the file to parse

Why can't this be automated?
```shell
curl ftp://ftp.unicode.org/Public/UNIDATA/UnicodeData.txt -o ./UnicodeData.txt
```

bd82:
> As for the lookahead function I've tried a few things without any real working solutions. Do you have any documentation for me to look out for without just going with an if and a few matches in the matchIdentifierJava ?

Not really this is a unique edge case...
Maybe leave more in-depth performance investigations for later after merging this PR.
The performance regression is not too bad and the source code of the custom token matcher should be
super easy to understand and maintain now 👍

Hawkurane:
Ok, sorry for not being able to answer lately. I'm gonna review the code tomorrow and force push on this PR with the new maintainable code. I'm gonna rewrite a few things as well to make it easier to understand and maintain. We can always leave the performance investigations for a future PR.

Hawkurane:
I've force pushed to squash the commits and changed the files. Could you check if everything is fine ? :blush:

packages/java-parser/src/tokens.js
```diff
@@ -32,17 +33,46 @@ FRAGMENT("HexDigits", "{{HexDigit}}(({{HexDigit}}|'_')*{{HexDigit}})?");
...
+
 const Identifier = createTokenOrg({
   name: "Identifier",
-  // TODO: Align with the spec, Consider generating the regExp for Identifier
-  //       as done in Esprima / Acorn
-  pattern: /[a-zA-Z_\\$][a-zA-Z_\\$0-9]*/
+  pattern: { exec: matchJavaIdentifier },
+  line_breaks: false,
+  start_chars_hint: Array.from(chars.firstIdentChar, x =>
+    String.fromCodePoint(x)
```
bd82: the start_chars_hints is utf16
see: https://sap.github.io/chevrotain/documentation/4_3_1/interfaces/itokenconfig.html#start_chars_hint

So I think you need to use String.fromCharCode

packages/java-parser/src/unicode.js
```diff
@@ -0,0 +1,144 @@
...
+  "Lo",
+  "Lt",
+  "Lu",
+  "Nl",
+  "Sc",
+  "Pc"
+]);
+
+// Adding remaining categories that can be used in a java identifier.
+// https://docs.oracle.com/javase/7/docs/api/java/lang/Character.html#isJavaIdentifierPart(char)
```
bd82: Should not we use Java 11 specifications? is there a difference?

packages/java-parser/src/unicode.js
```diff
@@ -0,0 +1,144 @@
...
+const restIdentUnicodeCategories = new Set(["Mn", "Cf"]);
+
+const manuallyAddedCharacters = new Set([]);
+// Below adding the isIdentifierIgnorable characters from Java Specs but not from FORMAT general category
+for (let i = 0; i < 8; i++) {
+  manuallyAddedCharacters.add(i);
+}
+for (let i = parseInt("000E", 16); i < parseInt("001B"); i++) {
```
bd82: Why do we need so many tiny for loops?
The only thing different between the loops is the different ranges.

```javascript
const ranges = [
  {start: " "000E", end: "001B"},
  {start: " "007F", end: "009F"},
// ...
]

ranges.forEach((range) => {
for (let i = parseInt(range.start, 16); i < parseInt(range.end); i++) {
  manuallyAddedCharacters.add(i);
}
})
```

packages/java-parser/src/unicode.js
```diff
@@ -0,0 +1,144 @@
...
+  }`;
+  fs.writeFileSync(path.resolve(__dirname, "unicodesets.js"), data, err => {
+    if (err) {
+      throw err;
+    }
+  });
+}
+
+// Calling the generation of the file.
+generateFile();
```
bd82: its wierd that reading the unicode is "inlined" in the source but writing the output
is wrapped in a function.

It would be cleaner if each step would be wrapped in a function so the end of the file would describe a **flow**:
```javascript

const unicodeData = readUnicodeData();
const {restCharCodes, firstCharCodes} = analyzeData(unicodeData);
generateJSUnicodeFile(restCharCodes, firstCharCodes);
```

packages/java-parser/src/tokens.js
```diff
@@ -32,17 +33,46 @@ FRAGMENT("HexDigits", "{{HexDigit}}(({{HexDigit}}|'_')*{{HexDigit}})?");
 FRAGMENT("FloatTypeSuffix", "[fFdD]");
 FRAGMENT("LineTerminator", "(\\x0A|(\\x0D(\\x0A)?))");
 
+function matchJavaIdentifier(char, startOffset) {
```
bd82: the first argument here is 'text' not 'char'

packages/java-parser/src/tokens.js
```diff
@@ -32,17 +33,46 @@ FRAGMENT("HexDigits", "{{HexDigit}}(({{HexDigit}}|'_')*{{HexDigit}})?");
...
+  while (chars.restIdentChar.has(charCode)) {
+    endOffset++;
+    charCode = char.codePointAt(endOffset);
+  }
+
+  // No match, must return null to conform with the RegExp.prototype.exec signature
+  if (endOffset === startOffset) {
+    return null;
+  }
+  const matchedString = char.substring(startOffset, endOffset);
```
bd82: I do not think the offsets are correct for unicode characters outside UTF16
Because:
```
"💩".length // === 2
```

bd82: If any of the Identifier characters are encoded using surrogate pairs we will need to take this into account.

bd82:
There has to be tests for such a complex scenario...

packages/java-parser/src/unicode.js
Hawkurane: No there isn't, for some reason java doc 11 has a different pathing in the link and I couldn't find it. I've fixed it now and pasted the link to java 11 docs.

Hawkurane:
As for the tests, what do you suggest? We could write a little program that uses such characters and parse it in a test file just to be sure.

the 💩 emoji is not parse-able in Java, but I've tried with another surrogate pair and it seems to work.
Also for some reason, parsing spring-boot I take only 7461ms.

Hawkurane:
I've automated the download and parsing of the unicode data file but there is a few problem that I think we should fix:
- The script downloads the file each time, if the file exists we shoudln't need to download it a n-th time. Downloading the file takes about 2 secs.
- We need to format:fix the unicodesets.js file, or the format:validate command will crash during ci tests. This command takes about 2 secs
- Currently downloading to a /tmp folder. If there is a future vscode plugin, this might cause problem to non-linux-like environments (Windows)

package.json
```diff
@@ -7,17 +7,14 @@
   "scripts": {
     "lerna:version": "lerna version",
     "lerna:publish": "lerna publish from-git --yes",
-    "ci": "npm-run-all download build lint format:validate test clean",
+    "ci": "npm-run-all build lint format:validate test",
```
bd82: Why do we avoid the downloads?

Hawkurane: I think downloading 2mb (supposed we don't download the license with it) each time we want to run the parser makes us lose a lot of time. Also, according to the [license](https://www.unicode.org/license.html), by downloading the file the user still accepts the license.
Also, as another solution we could just use another node_module to do the same work as we do. (i.e [npm unicode](https://github.com/eversport/node-unicodetable))
That is why I would either store it locally in the repo or use something else rather than downloading the file each time. Plus, according to the [Java Specs](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Character.html#isDefined(int)), it seems Java themselves use the UnicodeData.txt file

bd82: It is not each time we use the parser but each time we run the CI process.
And you could even do it so it would not run as part of the normal CI process 
but be a manual step that would be run manually only if there were changes to the generated .js file for the unicode data.

Just make sure that we comply with all the legal requirements if we include this as part of the repository and also filter it out of the npm package we are releasing (we do not need it at runtime).

Hawkurane: I think everything is fine as long as we issue a copy of the license and copyrights in the same folder as the UnicodeData.txt file (just as I did it). I'm thinking of changing the .html file to a .md file just so that we can remove the .css file to make it less intrusive in the project.

If we have the UnicodeData.txt file in the repo I don't think we need to deal with third party node_modules either.

Hawkurane:
Is everything fine with you? Should I squash the commits or is there something else we should review?

bd82:
You can squash and merge.
I am not a fan of keeping 3rd party code in the repo but, if you are fine with it...

Hawkurane:
I've squashed them.
I've let it into the repo but we can always change it to make it download in a later PR if people find it better to download it

bd82:
Just add an `.npmignore` file to exclude the resources folder from the npm package.
I'd rather not have end users download a large file that is not actually needed at runtime.

- https://docs.npmjs.com/files/package.json#files

after that you can merge this, you don't need another review from me 😄

Hawkurane:
I've added a .npmignore file, could you check just to be sure?
I don't have merge rights so you have to do it :+1:

.npmignore
```diff
@@ -0,0 +1 @@
+packages/java-parser/resources/Unicode/*
```
bd82: I think you want the .npmignore file right next to the relevant package.json
not at the root of the repository.

bd82:
>  I don't have merge rights so you have to do it 👍

Oh I thought you were part of the JHipster org.

Shaolans:
> Oh I thought you were part of the JHipster org.

We are :smile:, I think we have not introduced ourselves yet, we are new trainees under @jdubois supervision so we are still learning and we might need a guardrail to prevent any mistakes on jhipster projects.

bd82:
> We are 😄, I think we have not introduced ourselves yet, we are new trainees under @jdubois supervision so we are still learning and we might need a guardrail to prevent any mistakes on jhipster projects

I assumed as much, but then saw @Hawkurane  was not a member of the JHipster org on github.
Anyhow you cannot assume I will always have time to check things in a "live" fashion.
so perhaps a more async model would work better.
and you will also have to check each other's work when possible for Knowledge transfer 
and because I won't be able to review everything here 😄

Hawkurane:
I've changed the .npmignore's folder, everything should be fine. Do we put the unicodesets.js in the repo or does every end user have to parse the UnicodeData.txt file at least once before execution?

bd82:
> I've changed the .npmignore's folder, everything should be fine. Do we put the unicodesets.js in the repo or does every end user have to parse the UnicodeData.txt file at least once before execution?

If the unicodeData.txt is not located in the npm package how would the end user even be able to parse it?

and why should they parse it? would the result ever change for different end users?...

Hawkurane:
Currently the unicodesets.js file is not in the repo (it is the generated js file that is used by the tokens.js) and I fail to understand how end users can execute the parser when the unicodesets.js file is missing.
Should I have added it to the repo? I initially thought each end user had to parse the unicodedata.txt file at least once to generate the file so the parser could run properly. If we put the unicodesets.js file in the repo then I (finally) understand why we don't need the unicodedata.txt file in the npm package and that it is only used when maintaining the module

bd82:
> If we put the unicodesets.js file in the repo then I (finally) understand why we don't need the unicodedata.txt file in the npm package and that it is only used when maintaining the module

👍

Hawkurane:
I've added the generated file unicodesets.js to the repo ;)

bd82:
O.k. Lets merge this but there are a couple of changes that should be introduced afterwards in a separate commit

1. Mark the generated files clearly as generated (comment at start).
2. evaluate how to write the generated file in a way that it would take less space (it is 2mb+ now)
   - Is the set of firstChars strictly contained in nextChars? if so we can define nextChars set as first cloning firstChars set
  - Many of the characters appear in a range (without holes) we can write the ranges and loop over those
    to create the sets.
  - We could even write the values in base 16 to reduce the nuber of characters required.

Hawkurane:
Ok sure, I'll make an issue so we can keep tracking the things that needs to be done. It might be also good to recheck everything so we have a better view to what needs to be done and what priority each task has

bd82:
> Ok sure, I'll make an issue so we can keep tracking the things that needs to be done. It might be also good to recheck everything so we have a better view to what needs to be done and what priority each task has

Sounds good. Add another performance/optimizations pass to the list please 😄

bd82:
I've attempted to hack some additional performance for the unicode Identifier matching.
For example:
- explicit condition vs "true" literal

```javascript
// this means x is "truthy" or "==" (double equals)
// so could involve casting.
if (x) {

}

// this is faster
if (x === true) {

}
```

- Or reusing the same array instance (with different element) in the returned result.
   to avoid a new array initialization per each Identifier matched.

It seems these optimizations make too little difference as the main performance bottleneck
is during parsing, particularly the significant use of backtracking, so making the lexer x% faster 
does not matter much in the grand scheme of things if the parser if an order of magnitude slower.

PR #156, opened 2019-03-11
fix lexer: case where the lexer recognize OctalLiteral instead of Flo…

Shaolans:
…atLiteral

I was still trying to parse elasticsearch and I found another file that did not parse https://github.com/elastic/elasticsearch/blob/v6.6.1/server/src/test/java/org/elasticsearch/index/search/geo/GeoUtilsTests.java#L239.
The lexer is the culprit due to ```001.0``` to be recognized as 2 distinct tokens (001 as an octalLiteral and .0 as a FloatLiteral). Since the lexer choose the first match, a simple fix is to swap the order of creation between an OctalLiteral and FloatLiteral.
This bug should be the last one for this repository (#152 and #154).

### DIFF
--- a/packages/java-parser/src/tokens.js
+++ b/packages/java-parser/src/tokens.js
@@ -126,7 +126,6 @@ createToken({
   group: "comments"
 });
 createToken({ name: "BinaryLiteral", pattern: /0[bB][01]([01_]*[01])?[lL]?/ });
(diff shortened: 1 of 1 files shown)


bd82:
Good 👍 the longer alternative should appear first and the floatSuffix would ensure the two alternatives are always distinguishable.

PR #158, opened 2019-03-12
fix imports accept double semi-colon

Shaolans:
Hello, this time the parser failed with https://github.com/thinkaurelius/titan/blob/titan10/titan-all/src/main/java/com/thinkaurelius/titan/graphdb/tinkerpop/plugin/TitanGremlinPlugin.java#L12. This is due to the double semi-colon ```;;``` that is accepted by Java. I did an ad hoc fix that does not look pretty..., another solution would also be to add a big OPTION in "importDeclaration" rule.
But honestly I think ```;;``` is really ugly and I just want to PR the repository to delete the semi-colon :joy:.

### DIFF
--- a/packages/java-parser/src/productions/packages-and-modules.js
+++ b/packages/java-parser/src/productions/packages-and-modules.js
@@ -64,9 +64,6 @@ function defineRules($, t) {
     //                 distinguishing between the alternatives due to unbound common prefix.
     // TODO: A post parsing step is required to align with the official specs.
     //       The Identifier "var" is not allowed in all positions and variations of the importDeclaration
-    // Spec Deviation: The spec do not allow empty statement in between imports.
-    //                 However Java compiler consider empty statements valid, we chose
-    //                 to support that case, thus deviate from the spec.
     $.OR([
       {
         ALT: () => {
@@ -82,6 +79,10 @@ function defineRules($, t) {
           $.CONSUME(t.Semicolon);
         }
       },
+      // Spec Deviation: The spec do not allow empty statement in between imports.
+      //                 However Java compiler consider empty statements valid, we chose
+      //                 to support that case, thus deviate from the spec.
+      //                 See here: https://github.com/jhipster/prettier-java/pull/158
       {
         ALT: () => $.SUBRULE($.emptyStatement)
       }
--- a/packages/java-parser/src/productions/packages-and-modules.js
+++ b/packages/java-parser/src/productions/packages-and-modules.js
@@ -64,16 +64,28 @@ function defineRules($, t) {
     //                 distinguishing between the alternatives due to unbound common prefix.
     // TODO: A post parsing step is required to align with the official specs.
     //       The Identifier "var" is not allowed in all positions and variations of the importDeclaration
-    $.CONSUME(t.Import);
-    $.OPTION(() => {
-      $.CONSUME(t.Static);
-    });
-    $.SUBRULE($.packageOrTypeName);
-    $.OPTION2(() => {
-      $.CONSUME(t.Dot);
-      $.CONSUME(t.Star);
-    });
-    $.CONSUME(t.Semicolon);
+    // Spec Deviation: The spec do not allow empty statement in between imports.
+    //                 However Java compiler consider empty statements valid, we chose
+    //                 to support that case, thus deviate from the spec.
+    $.OR([
+      {
+        ALT: () => {
+          $.CONSUME(t.Import);
+          $.OPTION(() => {
+            $.CONSUME(t.Static);
+          });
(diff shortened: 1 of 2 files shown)


bd82:
Hi @Shaolans.

We need to fix the root issue...
This fix loses some **alignment with the specifications style** which is one of the main implementation concerns. There is also code duplication in this PR 😢 

- Does the specification allow multiple semicolons?

This seems none trivial, I don't understand for example why we have a GATE that looks exactly one token ahead (this would be done automatically by Chevrotain). So when we do have "root cause" fix ready
such a fix should be accompanied by a unit test.

See: https://github.com/jhipster/prettier-java/blob/master/packages/java-parser/test/bugs-spec.js

Shaolans:
Hello @bd82,
Yes you are right we need to fix the root issue, my bad ! I should have dig deeper.

> I don't understand for example why we have a GATE that looks exactly one token ahead (this would be done automatically by Chevrotain)

I don't know why I did that either, it is indeed not necessary.

> Does the specification allow multiple semicolons?
Yes the specifications does allow multiple semicolons, for instance inside a class declaration https://docs.oracle.com/javase/specs/jls/se11/html/jls-8.html#jls-ClassBody, but also with typeDeclaration https://docs.oracle.com/javase/specs/jls/se11/html/jls-7.html#jls-TypeDeclaration (that is why we can add semicolons right after the imports)
example:
```
import java.time.temporal.ChronoUnit;
import java.util.HashSet;
import java.util.Set;;;;;;;;
public class Test {}
;;;;;;
```
The parser works for this example since the semicolons are recognized as TypeDeclaration.

However, according to the specifications https://docs.oracle.com/javase/specs/jls/se11/html/jls-7.html#jls-OrdinaryCompilationUnit, the imports have to be declared one after the other. And every subrule possible for importDeclaration https://docs.oracle.com/javase/specs/jls/se11/html/jls-7.html#jls-ImportDeclaration "eat" at least an import token, thus not allowing multiple semicolons inside the "block" of imports.
So I think the parser follows the specifications but the Java compiler still accepts semicolons here 
(A similar problem will occurs if we add semicolons before the imports).
I am still not sure if this is really correct or did I missed something important ?

bd82:
I believe your understanding of the specifications is correct.

This raises the issue if we should comply with the Specifications or the official Java Parser.
I wonder what happens if that file with multiple semicolons at import section is opened in Eclipse/Intellij?
Would an error be shown? 

I honestly don't know if we should fix this or not.
But if we do choose to fix it I beleive the simplest solution would be to allow an **emptyStatement** as one of the alternatives of an **importDeclaration**.

https://docs.oracle.com/javase/specs/jls/se11/html/jls-14.html#jls-EmptyStatement

If we do this it needs to be clearly documented and combined with a unit test in 
https://github.com/jhipster/prettier-java/blob/master/packages/java-parser/test/bugs-spec.js
as it is such a strange edge case.

Shaolans:
>This raises the issue if we should comply with the Specifications or the official Java Parser.
I wonder what happens if that file with multiple semicolons at import section is opened in Eclipse/Intellij?
Would an error be shown?

With Intellij, it just gray out the semicolons and says that the semicolons are valid but useless.
```
Unnecessary semicolon ';' less... (Ctrl+F1) 
Inspection info: Reports on any unnecessary semicolons, whether between class members, inside block statements, or after class definitions. While valid Java, these semicolons are redundant, and may be removed
```

I don't know if we should fix it either, since that kind of case looks like a typo mistake...
I will leave the choice up to you.

bd82:
I think we should remove it.
But do so cleanly via adding another kind of alternative to the importDeclaration rule.

Shaolans:
I am not sure to understand, what should we remove ?

Shaolans:
I am not sure to understand what you mean by remove it, we let that case be or fix it as you said with an emptyStatement ?

bd82:
I don't know why I wrote "remove" I meant support the multiple semicolons 😄

Shaolans:
Oh right ! It does make a lot more sense :smiley:

packages/java-parser/src/productions/packages-and-modules.js
```diff
@@ -64,16 +64,28 @@ function defineRules($, t) {
...
+          });
+          $.SUBRULE($.packageOrTypeName);
+          $.OPTION2(() => {
+            $.CONSUME(t.Dot);
+            $.CONSUME(t.Star);
+          });
+          $.CONSUME(t.Semicolon);
+        }
+      },
+      {
```
bd82: The comment should be placed closer to the relevant alternative
and we should include a link to the discussion on this PR.

Shaolans: Got it :+1:

PR #162, opened 2019-03-20
Cst print

clement26695:
Implement the printers for a large part of the input. I leave the rest of the implementation for smaller PRs

### DIFF
--- a/packages/prettier-plugin-java/src/printers/types-values-and-variables.js
+++ b/packages/prettier-plugin-java/src/printers/types-values-and-variables.js
@@ -10,7 +10,11 @@ const {
   indent,
   dedent
 } = require("prettier").doc.builders;
-const { rejectAndJoin, rejectAndConcat } = require("./printer-utils");
+const {
+  rejectAndJoin,
+  rejectAndConcat,
+  sortClassTypeChildren
+} = require("./printer-utils");
 
 class TypesValuesAndVariablesPrettierVisitor {
   primitiveType(ctx) {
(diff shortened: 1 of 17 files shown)


packages/prettier-plugin-java/test/hello-world/_output.java
```diff
@@ -1,5 +1,46 @@
 package graphics.x.y.z;
-public class HelloWorld <AR> extends APZnd implements Leoq, LA {
-  public int t = expression;
-  public void test( int maz, final unannClassType go) throws testException block
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+public class HelloWorld <AR, AKDa> extends APZnd implements Leoq, LA {
```
bd82: We will need to split this to **multiple** smaller test cases

packages/prettier-plugin-java/src/printers/types-values-and-variables.js
```diff
@@ -115,26 +96,53 @@ class TypesValuesAndVariablesPrettierVisitor {
...
+      const startOffset2 = b.name
+        ? b.children.At[0].startOffset
+        : b.startOffset;
+      return startOffset1 - startOffset2;
+    });
+
+    const segments = [];
+    let currentSegment = [];
+
+    for (let i = 0; i < tokens.length; i++) {
```
bd82: should we use [].forEach or _.forEach

packages/prettier-plugin-java/src/printers/types-values-and-variables.js
```diff
@@ -46,65 +47,45 @@ class TypesValuesAndVariablesPrettierVisitor {
...
+    const tokens = sortClassTypeChildren(
+      ctx.annotation,
+      ctx.typeArguments,
+      ctx.Identifier
+    );
+
+    const segments = [];
+    let currentSegment = [];
+
+    for (let i = 0; i < tokens.length; i++) {
```
bd82: forEach?

bd82:
please squash this when merging

packages/prettier-plugin-java/test/hello-world/_output.java
clement26695: How do you think we should proceed ? One test for each printer function ?

bd82:
Note the travis voter is disabled. somebody with permissions needs to re-enable it 😄

PR #163, opened 2019-04-01
fix(annotations): fix annotations presentation

clement26695:


packages/prettier-plugin-java/test/hello-world/annotations_input.java
```diff
@@ -0,0 +1,9 @@
+package io.github.jhipster.sample;
+
+@SpringBootApplication
+@EnableConfigurationProperties({ LiquibaseProperties.class, ApplicationProperties.class })
+public class JhipsterSampleApplicationApp {
+
+    @PostConstruct @Override
+    public void initApplication() {}
+}
```
clement26695: Did you mean smaller test input like these, @bd82 ? I'm not sure on how to decompose the tests : do you think we should have a test per printer method ?

bd82: Have a look at the old prettier-java test cases
https://github.com/jhipster/prettier-java/tree/old-version/packages/prettier-plugin-java/tests

Issue #117, opened 2019-01-10
Attaching Comments to the CST


bd82:
The comments are currently collected inside a separate array of the lexing result.
That is not sufficient for the prettier scenario as the re-printed Java Text must include all comment
in the correct places.

What we need is to attach the comments to the nearest CST node
Or perhaps even node**s** (nearest before **and** after nodes).

This was done in the previous implementation by @thorbenvh8 
by adding logic to the basic parser [lookahead flows](https://github.com/jhipster/prettier-java/blob/old-version/packages/java-parser/src/parser.js#L2809-L2881).
and also by defining [many types of Comments variants](https://github.com/jhipster/prettier-java/blob/old-version/packages/java-parser/src/tokens.js#L58-L93).

My **naive** approach would be to first try to implement this as a **post processing** phase on the parsing results rather than inside the Parser itself. The reasons behind this desired approach are:
- that the comment attaching is not mandatory for all parsing flows, so it would nice for it to be easily toggle-able.
- Post processing such as building an AST from A CST is often an order of magnitude faster than the actual parsing, so another traversale of the CST and slight modifications for it do not concern me in terms of performance.
- The parser is already quite complex, I'd rather not introduce additional complexity into it and 
   have a **strong separation of concerns** between parsing and comments attaching topics.

bd82:
It may be worthwhile to investigate how @babel/parser attaches the comments
- https://babeljs.io/docs/en/next/babel-parser.html

Shaolans:
Hi @bd82,
I started to look into babel-parser to try to understand how comments attachment is done.
The [first version](https://github.com/babel/babel/issues/1486) directly use the attach comments functions from estraverse but they later implemented their own based on [it](https://github.com/estools/estraverse/blob/master/estraverse.js) for performance purposes.
I am checking on estraverse the attachComments function and it seems to traverse the AST and use the offsets to attach the comments, I am not really sure of it, I am still trying to figure out how exactly it is done.

Shaolans:
Hello,
I think I finally understood how the [function works](https://github.com/estools/estraverse/blob/master/estraverse.js) (still not quite sure if this is accurate thought). The first thing they do is to extends the range of comments (the startoffset and endoffset), it means that they "fill up" the gap of offsets of comments in between two tokens (gap made by spaces, carriage return...). I am not sure if I was clear, maybe a drawing is easier to understand.
![comments](https://github.com/Shaolans/Anything/blob/master/comments.jpg)
[...]

PR #165, opened 2019-04-01
add start and end offsets on nodes

Shaolans:
https://github.com/jhipster/prettier-java/issues/117#issuecomment-478562055

### DIFF
--- a/packages/java-parser/src/parser.js
+++ b/packages/java-parser/src/parser.js
@@ -148,9 +148,25 @@ class JavaParser extends Parser {
 
   // hack to turn off CST building side effects during backtracking
   // TODO: should be patched in Chevrotain
-  cstPostNonTerminal(ruleCstResult, ruleName) {
+  cstPostNonTerminal(node, ruleCstResult, ruleName) {
     if (this.isBackTracking() === false) {
-      super.cstPostNonTerminal(ruleCstResult, ruleName);
+      super.cstPostNonTerminal(node, ruleCstResult, ruleName);
+      this.addNodeOffsets(node);
+    }
+  }
+
+  addNodeOffsets(node) {
+    node.startOffset = Number.POSITIVE_INFINITY;
+    node.endOffset = Number.NEGATIVE_INFINITY;
+    for (const key in node.children) {
+      node.children[key].forEach(child => {
+        if (child.startOffset < node.startOffset) {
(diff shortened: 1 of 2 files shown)


packages/java-parser/src/parser.js
```diff
@@ -148,9 +148,25 @@ class JavaParser extends Parser {
...
-  cstPostNonTerminal(ruleCstResult, ruleName) {
+  cstPostNonTerminal(node, ruleCstResult, ruleName) {
     if (this.isBackTracking() === false) {
-      super.cstPostNonTerminal(ruleCstResult, ruleName);
+      super.cstPostNonTerminal(node, ruleCstResult, ruleName);
+      this.addNodeOffsets(node);
+    }
+  }
+
+  addNodeOffsets(node) {
```
bd82: This seems to only scan one level down, but it needs to be infinite levels downs to find all tokens...
I think it should be as a feature in Chevrotain see:
https://github.com/SAP/chevrotain/issues/932

Shaolans: > This seems to only scan one level down, but it needs to be infinite levels downs to find all tokens...

Since we are doing the process in the cstPost, I thought we have the children nodes built before the current one, so don't we just need to check one level down ? The only case that might be missing is the root of the cst since we are calling the rule as the entrypoint ? 
If we need to scan downs to find all tokens at each node, I think the performance might be affected.

> I think it should be as a feature in Chevrotain see:
> SAP/chevrotain#932

Sorry ! I have seen that thread.
So I should let chevrotain experts do it then.

bd82: > Since we are doing the process in the cstPost, I thought we have the children nodes built before the current one

Oops, yes you are correct 👍 

> So I should let chevrotain experts do it then.

Or contribute to chevrotain 😄

Shaolans: >  Or contribute to chevrotain :smile:

I would love to, but I think by the time I get the hang of Chevrotain project the feature will already be added to it :laughing:

bd82: I doubt it, It would have some added complexity (docs / additional perf concerns)
but the basic logic is isolated to a very small area.

Shaolans:
I am closing it since Chevrotain added the location features.

PR #166, opened 2019-04-02
add benchmark

### LINKED ISSUE #118: Parser Benchmark
The parse-samples script is used as a form of benchmark.
However a better benchmark could include comparison to other Parsers
(e.g Antlr4 Java  grammar / Original java-parser package based on peg.js)

Such a benchmark should also be implemented using benchmark.js 
instead of just "new Date.getTime()" counters.

Shaolans:
Here the PR to close #118, I am not quite sure if this is what you wanted.
One benchmark is comparing the parser in the current repository versus the the original parser (0.0.2) and the other benchmark is comparing between the parser in the master repository versus the local repository.


packages/java-parser/benchmark/benchmark.js
```diff
@@ -0,0 +1,62 @@
...
+  console.log(`updating prettier-java master`);
+  cp.execSync(`git pull origin master`);
+} else {
+  console.log(`cloning prettier-java`);
+  cp.execSync(`git clone https://github.com/jhipster/prettier-java`, {
+    stdio: [0, 1, 2]
+  });
+}
+
+const masterparser = require("./prettier-java/packages/java-parser/src/index");
```
bd82: i don't understand how would one of these paths point to the most recently release parser.

Shaolans: Oh, I thought that we were comparing the master repository on github and not the npm.
I will fix it.

Shaolans: It should be fixed now, it will point to the lastest parser available on npm.

packages/java-parser/benchmark/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "benchmark",
```
bd82: add "private":true

packages/java-parser/benchmark/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "url": "git+https://github.com/jhipster/prettier-java.git"
+  },
+  "author": "",
+  "license": "ISC",
+  "bugs": {
+    "url": "https://github.com/jhipster/prettier-java/issues"
+  },
+  "homepage": "https://github.com/jhipster/prettier-java#readme",
+  "dependencies": {
+    "npm-install-version": "^6.0.2"
```
bd82: nice tool

bd82:
So this looks good. 👍 

I would suggest moving it to a "**private**" dev-package managed by lerna.
Examples:
- https://github.com/SAP/chevrotain/tree/master/examples

Make sure to update the lerna.json file.
- https://github.com/SAP/chevrotain/blob/master/lerna.json

Shaolans:
This should be good now, I think

benchmark/java-parser/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "main": "benchmark.js",
+  "scripts": {
+    "start": "node benchmark.js"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/jhipster/prettier-java.git"
+  },
+  "author": "",
+  "license": "ISC",
```
bd82: license should be apache

bd82:
I don't think the package.lock file should be included as we are using yarn

Shaolans:
I made the changes, it should be good now.

PR #168, opened 2019-04-04
Add tests on stable reformatting

clement26695:


packages/prettier-plugin-java/scripts/single-printer-run.js
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable no-console */
```
bd82: what is the difference between this and the existing script: "update-test-output.js" ?

clement26695: It is almost the same. I wanted to be able to test just a sample, and not update all of the tests, but I didn't really thought about it.

I should just add an argument to the "update-test-output.js"  script to specify the folder I want to parse/print, or add a flag to use the single-printer-run folder

bd82: sounds good. Keep the default behavior of updating everything, that is a relevant
flow when we change the formatting of common elements.

clement26695: I did it like that : 

```js
const samplesDir =
  process.argv.indexOf("-single") > -1
    ? path.resolve(__dirname, "./single-printer-run")
    : path.resolve(__dirname, "../test");
```

I thought of doing something like that :
```js
const samplesDir =
  process.argv.indexOf("-d") > -1
    ? path.resolve(__dirname, process.argv[process.argv.indexOf("-d") + 1])
    : path.resolve(__dirname, "../test");
```
but I thought it could lead to big issue if someone put `node update-test-output.js "/"`

packages/prettier-plugin-java/scripts/update-test-output.js
```diff
@@ -5,7 +5,10 @@ const path = require("path");
 const fs = require("fs");
 const prettier = require("prettier");
 
-const samplesDir = path.resolve(__dirname, "../test");
+const samplesDir =
+  process.argv.indexOf("-single") > -1
+    ? path.resolve(__dirname, "./single-printer-run")
```
bd82: should we exclude ./single-printer-run folder from the source control using .gitignore

Issue #169, opened 2019-04-05
Improvement for generated files (unicodesets.js)


Hawkurane:
Right now, the unicodesets.js is a generated file by parsing the [UnicodeData.txt](ftp://ftp.unicode.org/Public/UNIDATA/UnicodeData.txt) file. Currently we are only generating the file for maintaining purpose while the end-users only use the generated file to prettify Java code. 

Here are a few things that can be done:

- [x] Mark the generated files clearly as generated (comment at start).
- [x] Evaluate how to write the generated file in a way that it would take less space (it is 2mb+ now)
- [x] Make sure the optimization and performance are still "good" if not better
Is the set of firstChars strictly contained in nextChars? if so we can define nextChars set as first cloning firstChars set

Many of the characters appear in a range (without holes) we can write the ranges and loop over those to create the sets.

We could even write the values in base 16 to reduce the number of characters required.

PR #170, opened 2019-04-05
Improve unicode generation

Shaolans:
This is an improvement of unicode generation from #169.
A short comment is added on the generated unicode file.
The size of the file is going from 2+mb down to ~200kb by reusing firstCharset and generating loop to write ranges.


Shaolans:
I extended the loop optimization, it is down to ~25kb now.

bd82:
This looks good 👍 

I will merge this:
A couple of small upgrades that can be done separately:
- Give meaningful names to variables and functions (we have a function called "f" here for some reason...).
- It's OK to generate a comment or there if it helps in understanding the file (imagine looking at it 6 months down the road when you don't have any context).
- What is the execution time of this large file compared to the previous one? do the loops increase it too much? if it is just a few mili-seconds it does not matter...

pascalgrimaud:
good job @Shaolans

Shaolans:
> Give meaningful names to variables and functions (we have a function called "f" here for some reason...).
> It's OK to generate a comment or there if it helps in understanding the file (imagine looking at it 6 months down the road when you don't have any context).

I called it "f", to gain some spaces but since I have refactored it to reduce the number of call, I will change the name.
For the comment, I was trying to save some space but know since we have a correct size file, I fill make a clear comment on it.

> What is the execution time of this large file compared to the previous one? do the loops increase it too much? if it is just a few mili-seconds it does not matter...

```
Unicode benchmark
Unicode without loop x 8.32 ops/sec ±1.33% (25 runs sampled)
Unicode with loops x 9.83 ops/sec ±0.88% (28 runs sampled)
Fastest is Unicode with loops
```
I made a little benchmark and it seems that the loops peforms event better !
Using loops seems to work better than loading a large size file :smile: 

> good job @Shaolans 
Thanks ! :laughing:

bd82:
> I called it "f", to gain some spaces but since I have refactored it to reduce the number of call, I will change the name.

Treat the generated code as regular code. The intent is not to reduce every single bit of space/memory.
Rather to make clear code that is easy to understand and that is in reasonable size.
If somebody needs to reduce the code size significantly they would use a code minifier (e.g uglifyJS).

> Unicode benchmark
Unicode without loop x 8.32 ops/sec ±1.33% (25 runs sampled)
Unicode with loops x 9.83 ops/sec ±0.88% (28 runs sampled)
Fastest is Unicode with loops

What is this benchmark? Parsing time or loading time for the unicode.js file?

Shaolans:
> What is this benchmark? Parsing time or loading time for the unicode.js file?
This is the benchmark of the loading time for unicodesets.js, the generated file.


The following one is benchmarking the time taken by unicode.js to generate unicodesets.js
```
Unicode benchmarking runtime unicode.js
Unicode without loop x 4.67 ops/sec ±0.89% (16 runs sampled)
Unicode with loops x 8.83 ops/sec ±0.73% (26 runs sampled)
Fastest is Unicode with loops
````

bd82:
> This is the benchmark of the loading time for unicodesets.js, the generated file.

Ohh, so 100ms to load that file? ouch

Shaolans:
> Ohh, so 100ms to load that file? ouch

It should be noted that I was using a pretty old computer and since I only wanted to see the difference I launch node command on unicodesets.js. It is probably better to just call require and clear the require cache at each run.

bd82:
I have a pretty modern computer and it was 50ms, after the upgrade to imperative logic
it is 30ms+ which is still a bit slow but not horrible... 😄

PR #171, opened 2019-04-06
unicode optimization: small upgrades

Shaolans:
Some change in the generated file in order to make it clear from #170

### DIFF
--- a/packages/java-parser/scripts/unicode.js
+++ b/packages/java-parser/scripts/unicode.js
@@ -161,10 +161,15 @@ function generateFile() {
   *   https://github.com/jhipster/prettier-java/pull/155
   */
   "use strict"
-  function addRanges(o, arrays) {
-    arrays.forEach(range => {
-      [...Array(range[1] - range[0] + 1).keys()].map(i => o.add(i + range[0]));
-    });
+  const addRanges = (set, rangesArr) => {
+    for (let i = 0; i < rangesArr.length; i++) {
+      const range = rangesArr[i];
+      const start =  range[0];
+      const end = range[1];
+      for (let codePoint = start; codePoint <= end; codePoint++) {
+        set.add(codePoint)
+      }
+    }
   };
   const fic = new Set([`;
   firstIdentCharCategories.forEach(el => {
--- a/packages/java-parser/scripts/unicode.js
+++ b/packages/java-parser/scripts/unicode.js
@@ -150,11 +150,20 @@ function readUnicodeData() {
 // Generating a unicodesets.js file so that we don't have to reparse the file each time the parser is called.
 function generateFile() {
   let data = `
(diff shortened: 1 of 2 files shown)


packages/java-parser/scripts/unicode.js
```diff
@@ -150,11 +150,20 @@ function readUnicodeData() {
...
+  * Java uses the same file UnicodeData.txt as the unicode.js script to define the unicodes.
+  * For more:
+  *   https://github.com/jhipster/prettier-java/issues/116
+  *   https://github.com/jhipster/prettier-java/pull/155
+  */
   "use strict"
-  const f = (o, a) => {
-    a.forEach(e => {
-      [...Array(e[1] - e[0] + 1).keys()].map(i => o.add(i + e[0]));
+  function addRanges(o, arrays) {
```
bd82: note that a simpler imperative approach would be much faster.
```javascript
const f = (set, rangesArr) => {
  for (let i = 0; i < rangesArr.length; i++) {
    const range = rangesArr[i];
    const start =  range[0];
    const end = range[1];
    for (let codePoint = start; codePoint <= end; codePoint++) {
      set.add(codePoint)
    }
  }
};
```

And arguably easier to understand as you are performing multiple transformations in a single line.
- https://youtu.be/ecekSCX3B4Q?t=1667

packages/java-parser/src/unicodesets.js
```diff
@@ -2236,7 +2245,7 @@ const mac_a = [
   [65056, 65071],
   [48, 57]
 ];
-f(ricd, mac_a);
+addRanges(ricd, mac_a);
 const ric = new Set(
```
bd82: combining the sets takes **half** the execution file of this script (once the range adding function is optimized).
Is there a faster way to do so?

packages/java-parser/scripts/unicode.js
Shaolans: I will keep that in mind :).

packages/java-parser/src/unicodesets.js
Shaolans: Since we are only calling "has" with the sets, what I am thinking of is replacing ```ric```, that "fake" a set like so:
```
const ric = {
  has: (elt) => {
    if(fic.has(elt)) {
      return true;
    }
    return ricd.has(elt);
  }
};
```

bd82: that may speed up the init time but at a cost of the actual runtime of parsing source code.

bd82:
I think you can merge this and consider future upgrades later.
Certain future upgrades may depend on implementation details of other optimizations
for example if we implement a faster identification of a simple (none unicode) identifier
that may affect how much we care about the runtime performance of the unicode identifier recognition as it is rare to use such unicode identifiers.

PR #183, opened 2019-04-11
attach comments to tokens

Shaolans:
I have added the comment attachment on tokens in the parser.
Since the cst printer is using the npm java-parser and not the one from the master, can we release a new version ? Since I need the comment attachment to add the support of comment in the cst printer.

### DIFF
--- a/packages/java-parser/src/index.js
+++ b/packages/java-parser/src/index.js
@@ -14,7 +14,6 @@ const BaseJavaCstVisitorWithDefaults = parser.getBaseCstVisitorConstructorWithDe
 function parse(inputText, entryPoint = "compilationUnit") {
   // Lex
   const lexResult = JavaLexer.tokenize(inputText);
-  parser.input = lexResult.tokens;
 
   if (lexResult.errors.length > 0) {
     const firstError = lexResult.errors[0];
@@ -27,6 +26,8 @@ function parse(inputText, entryPoint = "compilationUnit") {
         firstError.message
     );
   }
+
+  parser.input = attachComments(lexResult.tokens, lexResult.groups.comments);
 
   // Automatic CST created when parsing
   const cst = parser[entryPoint]();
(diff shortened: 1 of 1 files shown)


bd82:
> Since the cst printer is using the npm java-parser and not the one from the master, can we release a new version ?

It is not using anything from npm, you must not be using lerna properly...
Try running `yarn` at the top level of the project.

The internal package dependencies should be resolved via softlinks...

packages/java-parser/src/index.js
```diff
@@ -43,10 +46,71 @@ function parse(inputText, entryPoint = "compilationUnit") {
...
 }
 
+function attachComments(tokenArray) {
+  const tokens = tokenArray;
+  const newTokens = [];
+  const headcomments = [];
+  let i = 0;
+
+  // pushing comments at the beginning of the file to the leadingComments of the first non Comment token
+  while (tokenMatcher(tokens[i].tokenType, t.Comment) && i < tokens.length) {
```
bd82: would not it be simpler to collect the comments in a separate **group** and 
run over two indexes, one for the comments and one for the regular tokens?

https://sap.github.io/chevrotain/docs/features/token_grouping.html

There seems like a great deal of edge cases and conditionals in this logic, this can easily lead to bugs.

Shaolans: I guess we could do that way.

Shaolans: I pushed a new version of it, using the group comment.

bd82: i will have a look tomorrow probably

bd82:
I did not get around to reviewing it 😢 
But please do another pass on the logic, it seems strange we have many edge cases
but I suspect the logic of matching the indices should be more generic and not require many edge cases.

Shaolans:
Imo, we don't have a lot of edge cases. We only have two edge cases, one when the code starts with comments and one when the code ends with comments, they respectively attaches to the first (as leading) and to the last token (trailing).
The rest of the code is just searching for each comment where it should be placed and it attaches to the closest token. We also have two "if" just to check if we need the create an array  on a token to hold comments.

bd82:
o.k. than you can merge, I wanted to try simplifying this (not sure I would succeed...) But I do not seem to find the time for it...

Shaolans:
If I find a simplier way to do it, I will make a PR but for now I will merge this one.
By the way, we need to release a new version of the java-parser on npm since we need comments for the cst printer.

bd82:
>  By the way, we need to release a new version of the java-parser on npm since we need comments for the cst printer.

You need push permissions to master and execute this script to release a new version
- https://github.com/jhipster/prettier-java/blob/master/package.json#L9

Regardless of that I still think your dev env is not configured properly because you do not need to release
a new version of packageA to re-use it inside packageB in a mono repo...

PR #185, opened 2019-04-23
[WIP] Printing comments on the cstprinter

Shaolans:
Most of comments are printed, however it still needs some [improvements](https://github.com/jhipster/prettier-java/issues/117#issuecomment-484498134).
I am currently working on it.

### DIFF
--- a/packages/java-parser/src/index.js
+++ b/packages/java-parser/src/index.js
@@ -95,7 +95,7 @@ function attachComments(tokens, comments) {
 
     // attach comment to the closest token
     if (
-      element.startOffset - tokens[currentToken].endOffset <=
+      element.startOffset - tokens[currentToken].endOffset <
       tokens[currentToken + 1].startOffset - element.endOffset
     ) {
       if (!tokens[currentToken].hasOwnProperty("trailingComments")) {
--- a/packages/prettier-plugin-java/src/printers/arrays.js
+++ b/packages/prettier-plugin-java/src/printers/arrays.js
@@ -8,7 +8,11 @@ const {
   line,
   softline
 } = require("prettier").doc.builders;
-const { rejectAndConcat, rejectAndJoin } = require("./printer-utils");
+const {
+  rejectAndConcat,
+  rejectAndJoin,
+  getImageWithComments
+} = require("./printer-utils");
 
 class ArraysPrettierVisitor {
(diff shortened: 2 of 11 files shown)


bd82:
should the "getImageWithComments" be implicit or explicit?

Shaolans:
What do you mean by implicit or explicit ?

bd82:
This is explicit:
```javascript
    return group(
      rejectAndConcat([
        getImageWithComments(ctx.LCurly[0]),
        indent(
          rejectAndConcat([optionalVariableInitializerList, optionalComma])
        ),
        separator,
        getImageWithComments(ctx.RCurly[0])
      ])
    );
```

this is implicit:
```javascript
    return group(
      rejectAndConcat([
       ctx.LCurly[0],
        indent(
          rejectAndConcat([optionalVariableInitializerList, optionalComma])
        ),
        separator,
        ctx.RCurly[0]
      ])
    );
```

so the `getImageWithComments` is activated **implicitly**

Shaolans:
Most of utilities are from the Prettier API itself, I think only rejectAndConcat and rejectAndJoin are custom made function.

bd82:
> Most of utilities are from the Prettier API itself, I think only rejectAndConcat and rejectAndJoin are custom made function.

For discussion: Would it be better to wrap all the "builder" utilities with our own logic or explicitly call our logic in many many places?

Also consider we can add additional validation logic, e.g null/type checks for early failures.
- I personally had many cases where I had to debug code in prettier-plugin-toml because
of failures with a none relevant error message due to passing a null somewhere...

Shaolans:
I guess we could probably do something like this but for some cases we will still have to call the logic, for instance when we return directly the identifier image. Or when we have something like `int a,b,c` we wants spaces right after each comma.

Shaolans:
I am currently working on another branch with what you suggested with implicit [call](https://github.com/Shaolans/prettier-java/tree/comment_print_implicit).
This looks better I think.

bd82:
> I am currently working on another branch with what you suggested with implicit call.
This looks better I think.

Cool, that is just a suggestion, you should decide which is best 😄

Shaolans:
> Cool, that is just a suggestion, you should decide which is best 😄

Well, I have not thought about and I do think this is better that way 👍. I will probably close the PR and open another one.

PR #187, opened 2019-04-30
Printing comments on the cstprinter

### LINKED ISSUE #117: Attaching Comments to the CST

Shaolans:
Comment print with implicit call of get comments (previous discussion #185)
Fix #117


Shaolans:
Comment printing is pretty much done now, just eventually need some reviews before merging it.

packages/prettier-plugin-java/src/printers/prettier-builder.js
```diff
@@ -0,0 +1,103 @@
+"use strict";
+const prettier = require("prettier").doc.builders;
+
+function getImageWithComments(token, toadd = "") {
+  if (
+    !token.hasOwnProperty("leadingComments") &&
+    !token.hasOwnProperty("trailingComments")
+  ) {
+    return token.image + toadd;
+  }
```
clement26695: I'm not sure of the utility of this test ?

Why do we need the toadd parameter ?

packages/prettier-plugin-java/src/printers/printer-utils.js
```diff
@@ -1,10 +1,35 @@
...
-const { join, concat } = require("prettier").doc.builders;
+const {
+  isToken,
+  getImageWithComments,
+  join,
+  concat
+} = require("./prettier-builder");
+
+function buildFqn(tokens, dots) {
+  const images = tokens.map(tok => getImageWithComments(tok));
```
clement26695: Is this not handled in the join method ?

packages/prettier-plugin-java/src/printers/printer-utils.js
```diff
@@ -1,10 +1,35 @@
...
+  if (!Array.isArray(sepTokens)) {
+    return rejectAndJoin(sepTokens, elems);
+  }
+  const actualElements = reject(elems);
+  const res = [];
+
+  for (let i = 0; i < sepTokens.length; i++) {
+    res.push(
+      actualElements[i],
+      isToken(sepTokens) ? getImageWithComments(sepTokens[i]) : sepTokens[i]
```
clement26695: same

packages/prettier-plugin-java/src/printers/blocks-and-statements.js
```diff
@@ -56,7 +59,7 @@ class BlocksAndStatementPrettierVisitor {
       return this.visitSingle(ctx);
     }
 
-    return this.getSingle(ctx).image;
+    return getImageWithComments(this.getSingle(ctx));
```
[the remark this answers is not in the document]
Shaolans: I am not quite sure about that, since the purpose of this function is to just retrieve the single node existing node from a parent node. There is one case where we do not only want to get the token and we want to traverse the [nodes](https://github.com/jhipster/prettier-java/blob/master/packages/prettier-plugin-java/src/cst-printer.js#L61-L64)

packages/prettier-plugin-java/src/printers/prettier-builder.js
Shaolans: That's some old attempts, I will remove them.

packages/prettier-plugin-java/scripts/prettify-sample.js
```diff
@@ -0,0 +1,48 @@
...
+const prettier = require("prettier");
+
+const write = process.argv.indexOf("--write");
+
+if (process.argv.length < 3) {
+  console.log("node prettify-samples <path> [--write]");
+  process.exit(1);
+}
+const samplesDir = process.argv[2];
+const sampleFiles = klawSync(samplesDir, { nodir: true });
```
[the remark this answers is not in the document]
Shaolans: Yes it of course can :), done !

Shaolans:
@clement26695 Thanks for your review, I have taken account of all your suggestion and made the changes according to your feedback ;)

PR #190, opened 2019-05-07
Add test repository

### LINKED ISSUE #186: Reinforcing tests for the prettier-plugin-java package
Right now, we have a set of short samples and all it does is compare the output to an expected output. We should supplement additional tests to ensure the plugin still works as intended, especially as we are currently modifying it so it can print commentaries. Plus, it might be a problem for maintainability when we change the expected output in the future. 

Here a few ideas:

- [ ] Supplement the current samples with longer and better ones. With a lot of commentaries to make sure they are printed correctly.
- [ ] Find another way to test the prettier-plugin: prettify a repo that has correctly installed tests, and run them to make sure they are still functional
- [ ] ~Prettify a repo, remove all the whitespaces in both input and output and compare them to each other.~

Any ideas welcomed !

Shaolans:
I have added some repositories to test and it seems that the plugin can now successfully prettify spring-boot and still be semantically valid.
I also added some fixes.
fix #186

### DIFF
--- a/packages/prettier-plugin-java/scripts/update-test-output.js
+++ b/packages/prettier-plugin-java/scripts/update-test-output.js
@@ -10,9 +10,9 @@ let originalSamplesDir = samplesDir;
 if (process.argv.indexOf("-single") > -1) {
   samplesDir = path.resolve(__dirname, "./single-printer-run");
 } else if (process.argv.indexOf("-repository") > -1) {
-  const sample = path.resolve(__dirname, "../sample");
+  const testsamples = path.resolve(__dirname, "../test-samples");
   originalSamplesDir = path.resolve(__dirname, process.argv[3]);
-  samplesDir = path.resolve(sample, path.basename(originalSamplesDir));
+  samplesDir = path.resolve(testsamples, path.basename(originalSamplesDir));
   if (fs.existsSync(samplesDir)) {
     fs.removeSync(samplesDir);
   }
--- a/packages/prettier-plugin-java/scripts/update-test-output.js
+++ b/packages/prettier-plugin-java/scripts/update-test-output.js
@@ -10,9 +10,9 @@ let originalSamplesDir = samplesDir;
 if (process.argv.indexOf("-single") > -1) {
   samplesDir = path.resolve(__dirname, "./single-printer-run");
 } else if (process.argv.indexOf("-repository") > -1) {
-  const testsamples = path.resolve(__dirname, "../test-samples");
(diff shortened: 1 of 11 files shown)


bd82:
I think many test repos need to be removed.
Looks like the PR voters take ~2 hours now...
A voter should take at most 10mins.

https://travis-ci.org/jhipster/prettier-java/pull_requests

Shaolans:
Yes, that was what I was thinking too.
I will probably remove most of jhipster samples.

bd82:
It may also be possible to run some things in parallel, also trying out circle-ci may also be an option.
When I converted the Chevrotain project from travis to circle-ci it seems as though their containers were much faster...

Shaolans:
> It may also be possible to run some things in parallel
Do you mean running the java-parser and the prettier plugin on the separate jobs or running a job on each test repositories ?
By the way, with our current CI, we are running 3 jobs with node 8, 10 and 11. Is this necessary ?

> also trying out circle-ci may also be an option
I think we should migrate the CI to Azure Pipelines since the other repositories are migrated here.
And according @pascalgrimaud this is for: homogeneity for the JHipster orga and it is as fast as circle ci

bd82:
> Do you mean running the java-parser and the prettier plugin on the separate jobs or running a job on each test repositories ?

We could have 3 parallel jobs each with a different env variable and each doing a **different set** of repo checks.

>  By the way, with our current CI, we are running 3 jobs with node 8, 10 and 11. Is this necessary ?

We should change it to 8/10/12 and it is needed to ensure our code runs on all modern node.js versions.
- e.g using syntax not supported in old node.js versions.

> Btw, I removed most of repositories to just keep 2 JHipster repositories, Spring-Boot and java design patterns and the build time is now 11-13 minutes.

That is better, It is actually ~8 mins for the longest job so its perfectly fine.

packages/prettier-plugin-java/scripts/update-test-output.js
```diff
@@ -10,9 +10,9 @@ let originalSamplesDir = samplesDir;
 if (process.argv.indexOf("-single") > -1) {
   samplesDir = path.resolve(__dirname, "./single-printer-run");
 } else if (process.argv.indexOf("-repository") > -1) {
-  const sample = path.resolve(__dirname, "../sample");
+  const testsamples = path.resolve(__dirname, "../test-samples");
```
bd82: missing camel case (testSamples)

Issue #119, opened 2019-01-10
Cold Start Performance


bd82:
The Java Grammar is quite complex and Chevrotain performs many checks and analysis during
the initial Parser initialization.

The end result is that it takes > 100ms on a fast modern CPU to startup the parser.
Now this is not the worst thing in the world for a prettier-scenario, but it still warrants optimization.

The first thing that can be done is implement serialization and de-serialization of the grammar internal structure and inspect what potential performance benefits that can be gained.
- See: https://github.com/SAP/chevrotain/tree/master/examples/parser/serialized_grammar
- The complexity with this is to ensure the serialized artifacts are always up-todate
  and to provide a dev-mode where no serialized artifacts are needed to avoid slowing down the fast feedback cycles (which JavaScript excels in).

Additional options would be to investigate if anymore of the checks and analysis done during the Parser constructor can be avoided in productive flows, in other words the ability to have an even more **minimal startup** than just de-serializing a grammar.
- This could be more complex as it requires diving into Chevrotain internals.
- If this is possible it may lead to a Chevrotain generic feature that implement it.

The last possible optimization would be to actually pre-generate the lookahead functions
instead of creating them on the fly, But that is much farther into the future and outside the scope of prettier-java.

bd82:
The "Help Wanted" label refers mainly to trying to implement [grammar serialization](https://github.com/SAP/chevrotain/tree/master/examples/parser/serialized_grammar)
This is an existing and documented Chevrotain feature so knowledge of the internals is not needed.

Hawkurane:
I've tried to implement the grammar serialization and it seems using a custom token matcher (for java identifier) makes the pattern void (such as [this](https://github.com/Hawkurane/prettier-java/blob/2a9e12377fd098669ebb7a3aa5719bdc7420f5a9/packages/java-parser/src/gen/grammar.json#L2932)). This makes an issue when we are loading the grammar.json : the parser can't parse the identifiers anymore :cry:

bd82:
Interesting, this is strange because the Tokens themselves are not serialized, I will look into this

Hawkurane:
[...]

PR #194, opened 2019-05-09
Serialized grammar

Hawkurane:
Attempt at adding cold start performance enhancement

[issue](https://github.com/jhipster/prettier-java/issues/119)


bd82:
1. Is the gen directory part of the npm package?
2. What is the cold start performance improvement of this optimization? (we only want to add this complexity if it provides a nice boost)
3. Add a pre-commit hook to always re-generate the grammar on commits.
4. Add a test that ensures the generated grammar is up to date.

Hawkurane:
>  Is the gen directory part of the npm package?

As it is only generated when the grammar is being generated, it should be if we re-generate the grammar on commits

> What is the cold start performance improvement of this optimization? (we only want to add this complexity if it provides a nice boost)

After modifying a little bit the benchmark.js to compare it with the master version, about 1 bonus op/sec faster.

> Add a pre-commit hook to always re-generate the grammar on commits.

If it is worth it to implement the cold start performance, will do.

> Add a test that ensures the generated grammar is up to date.

Are the CI tests not enough?

bd82:
> After modifying a little bit the benchmark.js to compare it with the master version, about 1 bonus op/sec faster.

A separate test is needed for clod starting. one that only initializes the Parser...

bd82:
> Are the CI tests not enough?

The tests passing does not necessarily mean that the serialized grammar is up to date.

Hawkurane:
How can we test if the serialized grammar is up to date?
Supposed the pre commit hooks are buggy / removed in the future for reasons.

bd82:
Consider that the pre-commit hooks are not guaranteed to run.
For example if someone modifies source code directly via github web page.
Or if someone checksout the repo and modifies source code without installing the npm dependencies first.

You can test that the generated code is up to date by re-generating it and comparing the newly generated sources and the ones currently on the git HEAD.

See:
https://github.com/SAP/ui5-typescript/blob/master/packages/ts-types/test/snapshots.spec.js

Hawkurane:
> A separate benchmarking is needed for cold starting. one that only initializes the Parser...

I have been trying to implement a cold start performance benchmark but I am quite unsure what to bench again:

Do I compare the same parser initialization with and without already generated grammar?
Do I compare npm or/and master version with current local version with and without already generated grammar?

bd82:
> I have been trying to implement a cold start performance benchmark but I am quite unsure what to bench again:

> Do I compare the same parser initialization with and without already generated grammar?
Do I compare npm or/and master version with current local version with and without already generated grammar?

I would compare local version with npm version.
In our case this is less useful because of rapid development and not many npm versions released.
But it should be easy to modify the benchmark (locally) to compare two local versions (or pass a flag to enable/disable serialization)

Hawkurane:
It seems the local parser is faster when it comes to only initializing (38.17 vs 43.73 ops/sec).

Could you review my files if you have the time? @bd82

package.json
```diff
@@ -8,18 +8,19 @@
...
-    "build": "node packages/java-parser/scripts/unicode.js packages/java-parser/resources/Unicode/UnicodeData.txt; prettier --write packages/java-parser/src/unicodesets.js; node packages/java-parser/scripts/gen-diagrams.js",
+    "format:fix-generated": "prettier --write packages/java-parser/src/unicodesets.js packages/java-parser/src/gen/grammar.json;",
+    "build": "node packages/java-parser/scripts/unicode.js packages/java-parser/resources/Unicode/UnicodeData.txt; node packages/java-parser/scripts/gen-diagrams.js",
     "prettier": "prettier --plugin=./packages/prettier-plugin-java --parser=java",
     "update-test-outputs": "node packages/prettier-plugin-java/scripts/update-test-outputs.js"
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged"
+      "pre-commit": "lint-staged && node packages/java-parser/scripts/gen-grammar.js && git add packages/java-parser/src/gen/grammar.json"
```
bd82: I am thinking that maybe we should not commit the generated serialized grammar json.
What do you think?

bd82: We probably want a script to generate the serialized format in package json
and call it in the hook as "npm run SCRIPT_NAME"

bd82:
> It seems the local parser is faster when it comes to only initializing (38.17 vs 43.73 ops/sec).

Does this mean we are only looking at a 10% speed boost by activating the serialization?

packages/java-parser/src/index.js
```diff
@@ -2,8 +2,20 @@
...
+function initializeParser() {
+  let serializedGrammar;
+  try {
+    serializedGrammar = require("./gen/grammar.json");
+  } catch (err) {
+    //
+  }
+  const parser = new JavaParser(serializedGrammar);
+  //generate the grammar.
+  require("../scripts/gen-grammar");
```
bd82: why do we regenerate the serialized format on parser initialization ?
Would initializing the parser change the grammar?

Hawkurane: sorry, it seems by refactoring my code i removed a feature, but originally I meant to only generate the grammar if it didn't find an initial grammar.json file

package.json
Hawkurane: I think we should stay consistent, we are already committing a generated file (unicodesets.js) and using a script to maintain every once in a while.
Imo, we should commit the grammar.json file and use the script to serialize the grammar to maintain. I'm unsure whether we shoud regenerate it. It generates a new grammar at each pre-commit but I don't think it adds it to the commit at the moment: right now it's mostly used for local development.

Hawkurane:
> Does this mean we are only looking at a 10% speed boost by activating the serialization?

If I didn't mistake benchmark.js' inner workings, we are looking at a 10% speed boost in initializing the parser for activating the serialization.

packages/java-parser/src/index.js
bd82: Did you re-benchmark after removing this redundant init?

Hawkurane: I'm not using the same computer but I get NPM 41.35 ops/sec vs Local 45.78 ops/sec

bd82:
So maybe the benefits of this optimization are not worth the complexity. 😢

Hawkurane:
I don't think the optimizations are worth the complexity at all, should we close the issue and the PR?

bd82:
> I don't think the optimizations are worth the complexity at all, should we close the issue and the PR?

I agree,

PR #196, opened 2019-05-10
Remove empty statement

clement26695:


packages/prettier-plugin-java/test/comments/comments-blocks-and-statements/_output.java
```diff
@@ -3,7 +3,7 @@
 
   public void myFunction(int arg1) {
     try {
-      ;// Empty Statement
+      // Empty Statement
```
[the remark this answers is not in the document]
bd82: I am with @Shaolans here. I think this represents several unique edge cases.
I would like to see most of those tested, perhaps even with a separate test input/output pair.

edge cases:
- empty statement with comment.
- empty statement in for/do header.
- empty statement without comment

clement26695:
I added some tests and refactored the code to not modify the CST during processing.

There is still an issue in the tests I need to solve (issue on stable reformating)

packages/prettier-plugin-java/src/printers/printer-utils.js
```diff
@@ -156,8 +156,44 @@ function findDeepElementInPartsArray(item, elt) {
...
+function hasTrailingComments(token) {
+  return token.trailingComments !== undefined;
+}
+
+function hasComments(token) {
+  return hasLeadingComments(token) || hasTrailingComments(token);
+}
+
+function isExplicitLambdaParameter(ctx) {
+  return (
```
bd82: JavaScript needs Kotlin's null safety features 😄 
https://kotlinlang.org/docs/reference/null-safety.html#safe-calls

packages/prettier-plugin-java/test/empty_statement/_input.java
```diff
@@ -1,5 +1,27 @@
...
+  public void emptyStatementWithComment() {
+    ;;//EmptyStatement
+  }
+
+  public void simpleForWithEmptyStatement() {
+    for (;;);
+
+    for (;;)/*test*/;
+
+    for (;;);/*test*/
```
bd82: worth checking edge case of comments _between_ the for loop's semi colons?

clement26695: These semi colons are not considered as empty statements, so that's fine, IMO. But I can add some, that can not harm !

Issue #195, opened 2019-05-10
Setting up prettier-java for Azure Pipelines


Shaolans:
Since we will migrate our CI from Travis CI to Azure Pipelines, we need to set up the project in the JHipster organization on Azure. :smile:

pascalgrimaud:
Thanks. I'll do it asap

bd82:
Cool I would be interested in seeing a working example of Azure Pipelines.
I've only used circle-ci/Travis so far.

PR #197, opened 2019-05-13
Refactoring Tests

Shaolans:
I have made some refactoring of the tests and added the azure pipelines configuration file.
It launches 3 tests of the repository by changing the node versions and 2 tests are testing repositories only (node 10).
You can check here, on my fork the [builds](https://dev.azure.com/lingchun/prettier-java/_build?definitionId=2)

#195


bd82:
Can we do the AzurePipelines and the refactoring of the tests in separate PRs?

Shaolans:
Well, the AzurePipelines is just contained in [azure-pipelines.yml](https://github.com/jhipster/prettier-java/pull/197/files#diff-fec826feae04e51c0d94076385408bdc) and since I needed the refactoring for the CI I did both at the same time, is this that troublesome ?

azure-pipelines.yml
```diff
@@ -0,0 +1,63 @@
...
+      node_10:
+        node_version: '10'
+        TEST_REPOSITORY: 'CORE'
+      node_8:
+        node_version: '8'
+        TEST_REPOSITORY: 'CORE'
+  variables:
+    TEST_REPOSITORY: $(test_repository)
+  pool:
+    vmImage: 'ubuntu-latest'
```
bd82: Can we used a fixed version so the CI would be reproducible ?

bd82:
> since I needed the refactoring for the CI I did both at the same time, is this that troublesome ?

If you need both for this to work its o.k.
Otherwise it would a case of messy commit history without cause.

package.json
```diff
@@ -10,6 +10,9 @@
     "lerna:publish": "lerna publish from-git --yes",
     "ci": "npm-run-all build lint format:validate test",
     "test": "lerna run test --stream",
+    "test:java-parser": "mocha packages/java-parser/test/*.js",
```
bd82: it is better if the upper level package.json would not include specific
instructions for scripts belonging to inner packages.

azure-pipelines.yml
```diff
@@ -0,0 +1,63 @@
...
+        versionSpec: $(node_version)
+      displayName: 'Install Node.js'
+    - script: |
+        yarn
+      displayName: 'Install dependencies'
+    - script: |
+        yarn run ci
+      displayName: 'Run CI'
+
+- job: Extended_tests
```
bd82: "extended_tests" --> "e2e tests" ?

azure-pipelines.yml
```diff
@@ -0,0 +1,63 @@
...
+  strategy:
+    matrix:
+      jhipster-1:
+        TEST_REPOSITORY: 'JHIPSTER-1'
```
bd82: can we reference this variable in the script command below.
So instead of using an env variable we would pass the name of the yarn script to run?

So we would execute:
- lerna run ${SCRIPT_NAME}

and the value would be "e2e-A" or "e2e-B" for example.

I prefer to avoid depending env variables because it makes it harder to execute and reproduce the build locally.

bd82:
Should not we activate azure pipelines before we can merge this?

pascalgrimaud:
@bd82 : yes, it's on my todo list. I'll do it soon

azure-pipelines.yml
Shaolans: I will rename it "e2e_tests" since we can add spaces on the job name

azure-pipelines.yml
Shaolans: Yes, of course, I will add 12.2.0, 10.15.3 and 8.16.0

package.json
Shaolans: I will remove them, I just realised that we can use scope with lerna

azure-pipelines.yml
Shaolans: I have made the modifications to not use env variables.

Shaolans:
@bd82 I have made the changes, what do you think now ?

> If you need both for this to work its o.k.
> Otherwise it would a case of messy commit history without cause.

When everything will be ok to merge, I can just remove the azure-pipelines.yml and then merge it and create a new PR for it.

bd82:
> If this is now good to merge, I can just remove the azure-pipelines.yml and then merge it and create a new PR for it.

Sounds like a plan, that way it would be easier to review the Azure pipeline changes separately.

PR #198, opened 2019-05-14
Fix multi-lines comments

clement26695:


packages/prettier-plugin-java/test/comments/class/_output.java
```diff
@@ -351,12 +351,9 @@ public V set(int rowIndex, int columnIndex, @Nullable V value) {
...
   * @param valueClass class of values stored in the returned array
   */
-  @GwtIncompatible// reflection
-
+  @GwtIncompatible // reflection
   public V[][] toArray(Class<V> valueClass) {
-    @SuppressWarnings("unchecked"
-    )// TODO: safe?
-     V[][] copy = (V[][]) Array.newInstance(
+    @SuppressWarnings("unchecked") V[][] copy = (V[][]) Array.newInstance( // TODO: safe?
```
Shaolans: I don't know if we should keep annotation and statement in the same line (when there is only one annotation)

Hawkurane: The only moment we would to keep it inline is when we attach annotation to function parameters imo.

e.g
`public void f(@NotNull int x){ ... }`

clement26695: Ok, but it should be done in another PR, don't you think ?

PR #200, opened 2019-05-16
Azure pipelines

Shaolans:
Here is the Azure Pipelines configuration. It launches the test for node 8.16.0, 10.15.3 and 12.2.0 and launches extended end to end test on jhipster repositories.

related discussions #197

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -28,7 +28,7 @@ jobs:
       node_8:
         node_version: '8.16.0'
   pool:
-    vmImage: 'ubuntu-latest'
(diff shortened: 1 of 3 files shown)


bd82:
This looks good. 👍 

Please remove the travis build file.

pascalgrimaud:
@bd82 : I'd prefer to keep Travis for now, to be sure Azure works well here. Then we can remove it later. We shouldn't go too fast :)

bd82:
> @bd82 : I'd prefer to keep Travis for now, to be sure Azure works well here. Then we can remove it later. We shouldn't go too fast :)

o.k.

Issue #113, opened 2019-01-05
Parser Gaps


bd82:
- [x] Bit shifting operators must **not** include whitespace
       https://github.com/jhipster/prettier-java/blob/5ef66ae309b41e6634d6e027631c824482d4af96/packages/java-parser/src/productions/expressions.js#L150-L156
- [ ] Some optimized backtracking logic do not provide location information on error, e.g:
      https://github.com/jhipster/prettier-java/blob/5ef66ae309b41e6634d6e027631c824482d4af96/packages/java-parser/src/productions/classes.js#L784-L787
- [x] Some matches vs Identifiers do not take token categories into account, the chevrotain.tokenMatcher must be used instead., e.g: https://github.com/jhipster/prettier-java/blob/5ef66ae309b41e6634d6e027631c824482d4af96/packages/java-parser/src/productions/classes.js#L731
- [ ] Need farther investigation on reducing the use of backtracking.
- [ ] The var keyword is only allowed to be used as an Identifier in certain contexts and disallowed in others, this is currently not implemented by the parser.
  - e.g: https://docs.oracle.com/javase/specs/jls/se11/html/jls-3.html#jls-TypeIdentifier
- [x] Attaching Comments to the CST (#117)

pascalgrimaud:
Adding a bounty on this

bd82:
Not sure this is suitable for a bounty, it is more like a list of issues...

pascalgrimaud:
Ok, just tell me where I can reassign the bounty

bd82:
>  The var keyword is only allowed to be used as an Identifier in certain contexts and disallowed in others, this is currently not implemented by the parser.
e.g: docs.oracle.com/javase/specs/jls/se11/html/jls-3.html#jls-TypeIdentifier

This may be a single issue. Which may be important enough for bounty as it touches on correctness.

> Some optimized backtracking logic do not provide location information on error, e.g:

This is a separate issue, which is not specific enough as it does not specify the list
of such errors that is expected to include location information. It may be less important as it deals with performance and simplification.

> Need farther investigation on reducing the use of backtracking.

This is another separate issue and it may be completely "theoretical" So I am not sure it is suitable for bounty.

clement26695:
I agree. We should write another issue for the var/TypeIdentifier issue. It is not prioritary for the moment as it does not cause any bug, but it should be dealt with at some point

PR #202, opened 2019-05-20
Handle blank lines

clement26695:
To test it:
```bash
git clone https://github.com/clement26695/chevrotain.git
cd chevrotain/packages/chevrotain
git checkout add-location-information
yarn link
```
Then, go to the prettier-java repo and
```bash
cd packages/java-parser
yarn link "chevrotain"
```


Shaolans:
This looks good to me :+1:

bd82:
Try using Chevrotain 4.7.0, the cstNodes location tracking feature is now available.
- https://github.com/SAP/chevrotain/issues/932

Not the configuration property is called nodeLocationTracking and not nodePositionTracking

clement26695:
Done @bd82 ! 

I had to ignore a few more issues with Chevrotain to make it work, but it seems ok.

bd82:
> I had to ignore a few more issues with Chevrotain to make it work, but it seems ok.

What issues?

packages/java-parser/src/parser.js
```diff
@@ -42,6 +42,18 @@ class JavaParser extends Parser {
...
+        },
+        lambdaParameterType: {
+          OR: true
+        },
+        annotation: {
+          OR: true
+        },
+        localVariableType: {
+          OR: true
+        },
```
clement26695: These ones : binaryExpression, lambdaParameterType, annotation and localVariableTyp

bd82: That is probably due to [changes in 4.5.0](https://sap.github.io/chevrotain/docs/changes/CHANGELOG.html#_4-5-0-5-25-2019) for better detecting of ambiguities related to token categories.

Make sure that when you add new ignores you are not hiding a bug.
Most Chevrotain parsers should not use these "ignoredIssues" but the Java Parser is different because of the use of backtracking. So make sure all the new "ignoredIssues" are actually  resolved by backtracking and not a true problem.

Shaolans: annotation: we are using backtracking to resolve ambiguities
localVariableType and lambdaParameterType: same prefix of var for the var token but also unnannoted class type but this is a known issue about special handling of keywords https://github.com/jhipster/prettier-java/issues/113.
binaryExpression: "<" and ">" ambiguities but I think this is handled by looking token ahead.
Another ambiguities because of instanceof and BinaryOperator and BinaryOperator is a category of instanceof, I do not think this is an issue as we are trying to match the token instanceof first (and the following subrule are different I think)

PR #203, opened 2019-05-24
update README: add azure badge

Shaolans:


Shaolans:
I will add it while updating the README

PR #204, opened 2019-05-24
Update readme

### LINKED ISSUE #192: Update README.md files
I think it is time to update the main and child readme.md files.
e.g:
- There is no readme.md for prettier-plugin-java
  - https://github.com/jhipster/prettier-java/tree/master/packages/prettier-plugin-java

- The root readme roadmap is probably mostly done or irrelevant by now.

- The root readme should link to the sub packages of this mono repo.

Shaolans:
I updated the README and added one for the prettier-plugin-java, but it will only work on the next release (on npm we still have the old one)

Fix #192

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,12 @@
     :construction: Work in Progress! :construction:
 </p>
 
-[![Build Status](https://travis-ci.org/jhipster/prettier-java.svg?branch=master)](https://travis-ci.org/jhipster/prettier-java)
+[![Azure DevOps Build Status][azure-devops-image]][azure-devops-url-main] [![Build Status][travis-image]][travis-url]
+
+[azure-devops-image]: https://dev.azure.com/jhipster/prettier-java/_apis/build/status/jhipster.prettier-java?branchName=master
+[azure-devops-url-main]: https://dev.azure.com/jhipster/prettier-java/_build
+[travis-image]: https://travis-ci.org/jhipster/prettier-java.svg?branch=master
+[travis-url]: https://travis-ci.org/jhipster/prettier-java
 
 # Prettier Java
 
--- a/README.md
+++ b/README.md
@@ -41,6 +41,21 @@ This project contains 2 packages:
 
 This need to updated when the next version will be released.
 
+In the meantime, you can run the plugin by following these steps:
+
+```
+git clone https://github.com/jhipster/prettier-java
+cd prettier-java
+yarn
+cd packages/prettier-plugin-java/scripts
+node update-test-output.js -single
+```
+
+It will format the code contained in `packages/prettier-plugin-java/scripts/single-printer-run/_input.java` and ouput the formatted code in `packages/prettier-plugin-java/scripts/single-printer-run/_output_.java`.
+
+It is also possible to format a folder with the following command:
+`node update-test-output.js -repository path-to-folder`. The output will be stored in `packages/prettier-plugin-java/test-samples/`.
+
 ## Contributing
 
 Contributions are very welcome.
--- a/README.md
+++ b/README.md
@@ -29,8 +29,19 @@ It could even be used inside a browser.
 
 This project contains 2 packages:
 
+- [prettier-plugin-java](./packages/prettier-plugin-java) A plugin for
+  Prettier to format Java code
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -20,26 +25,36 @@ What this means is that unlike many other Prettier plugins,
 `prettier-java` has **no additional runtime pre-requisites** (e.g: Python executable).
 It could even be used inside a browser.
 
+## Subpackages
+
+This project contains 2 packages:
```
bd82: May be worth experimenting with add npm icon links next to each package name.
You can get those links/imgs from: https://shields.io/category/version

README.md
```diff
@@ -20,26 +25,36 @@ What this means is that unlike many other Prettier plugins,
...
+
+- [java-parser](./packages/java-parser) A Java Parser using [Chevrotain](https://github.com/SAP/chevrotain) which output a **C**oncrete **S**yntax **T**ree
+- [prettier-plugin-java](./packages/prettier-plugin-java) A plugin for Prettier to format Java code
```
bd82: List the prettier-plugin-java package first.
Most people who come to this repo are interested in the prettier plugin not its implementation details

packages/prettier-plugin-java/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+The plugin implementation is pretty straightforward as it uses [java-parser](../java-parser) (thanks to Chevrotain) visitor to traverse the **C**oncrete **S**yntax **T**ree and apply the format processing on each node (it uses Prettier API).
+
+## Installation
+
+### Pre-requirements
+
+Since the plugin is meant to be used with Prettier, you need to install it:
+
+`npm install prettier --save-dev`
```
bd82: it is best to install prettier with exact versions.
I think it is :
`--exact`

See: https://prettier.io/docs/en/install.html

packages/prettier-plugin-java/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+`yarn add prettier-plugin-java --dev`
+
+### CLI
+
+If you installed Prettier globally and want to format java code via the CLI, run the following command:
+
+`npm install -g prettier-plugin-java`
+
+The plugin with be automatically loaded, check [here](https://prettier.io/docs/en/plugins.html#using-plugins) for more.
```
bd82: with -> will

packages/prettier-plugin-java/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Usage
```
bd82: I think the main usage would be via CLI not via programmatic API.

packages/prettier-plugin-java/README.md
```diff
@@ -0,0 +1,111 @@
...
+  tabWidth: 2
+});
+```
+
+## Example of formatted code
+
+### Input
+
+```java
+public class HelloWorld {
```
n-peugnet: The Input seems to be the same as the output in this example.
However thank you for this, I will give it a go this evening

bd82:
We can add a "playground" using gitpod once the npm package is released.
Or maybe it is possible to do it even earlier without needing the new npm package.

See how I've done it in the toml prettier plugin:
- https://github.com/bd82/toml-tools/tree/master/packages/prettier-plugin-toml#try-it-out

packages/prettier-plugin-java/README.md
Shaolans: I will just invert those two then.

packages/prettier-plugin-java/README.md
Shaolans: I made a mistake here thanks !

Shaolans:
I made the changes :+1:

> We can add a "playground" using gitpod once the npm package is released.

I will look into it later :smile:

packages/prettier-plugin-java/README.md
n-peugnet: Sorry if I am being rude, but it doesn't look like it has changed

Shaolans: No not at all :smile:, thanks to point that again, this is due to one of git hook...

Shaolans: I linked the java plugin to prettier and forgot to remove it and since we run prettier on commit, it formatted the java code inside the markdown... Thanks :+1:

PR #208, opened 2019-06-08
Update lerna, husky, eslint & lint-staged

clement26695:


bd82:
Try enabling greenkeeper or dependentBot and it will help keep the depedencies up-to-date in a more automated manner. 

e.g: https://github.com/SAP/chevrotain/pull/971

bd82:
I see dependent-bot is already enabled on this repo.
Why do we need manual deps updates in this case?

clement26695:
I'm not sure why, but it doesn't update package.json, but only the yarn.lock file

bd82:
> I'm not sure why, but it doesn't update package.json, but only the yarn.lock file

Wierd maybe it only supports updating according to version ranges...

PR #213, opened 2019-06-13
Aligned string literals to java spec

Hawkurane:
This is a PR to align the String Literal definition to the Java Spec. I've reused the CharLiteral regex as it seems to be in accordance to the Java Specs

### DIFF
--- a/packages/java-parser/src/tokens.js
+++ b/packages/java-parser/src/tokens.js
@@ -203,8 +203,7 @@ createToken({
 });
 createToken({
   name: "StringLiteral",
-  // TODO: align with the spec, the pattern below is incorrect
-  pattern: /"[^"\\]*(\\.[^"\\]*)*"/
+  pattern: /"(?:[^\\"]|\\(?:(?:[btnfr"'\\/]|[0-7]|[0-7]{2}|[0-3][0-7]{2})|u[0-9a-fA-F]{4}))*"/
 });
 
 // https://docs.oracle.com/javase/specs/jls/se11/html/jls-3.html#jls-3.9
--- a/packages/java-parser/src/tokens.js
+++ b/packages/java-parser/src/tokens.js
@@ -40,9 +40,9 @@ FRAGMENT("HexDigits", "{{HexDigit}}(({{HexDigit}}|'_')*{{HexDigit}})?");
 FRAGMENT("FloatTypeSuffix", "[fFdD]");
 FRAGMENT("LineTerminator", "(\\x0A|(\\x0D(\\x0A)?))");
 FRAGMENT("UnicodeMarker", "uu*");
-FRAGMENT("UnicodeEscape", "\\{{UnicodeMarker}}{{HexDigit}}{4}");
-FRAGMENT("RawInputCharacter", "\\{{UnicodeMarker}}[0-9a-fA-F]{4}");
-FRAGMENT("UnicodeInputCharacter", "[{{UnicodeEscape}}{{RawInputCharacter}}]");
+FRAGMENT("UnicodeEscape", "\\\\{{UnicodeMarker}}{{HexDigit}}{4}");
+FRAGMENT("RawInputCharacter", "\\\\{{UnicodeMarker}}[0-9a-fA-F]{4}");
+FRAGMENT("UnicodeInputCharacter", "({{UnicodeEscape}}|{{RawInputCharacter}})");
 FRAGMENT("OctalDigit", "[0-7]");
 FRAGMENT("ZeroToThree", "[0-3]");
 FRAGMENT(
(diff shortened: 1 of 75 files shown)


packages/java-parser/src/tokens.js
```diff
@@ -203,8 +203,7 @@ createToken({
 });
 createToken({
   name: "StringLiteral",
-  // TODO: align with the spec, the pattern below is incorrect
-  pattern: /"[^"\\]*(\\.[^"\\]*)*"/
+  pattern: /"(?:[^\\"]|\\(?:(?:[btnfr"'\\/]|[0-7]|[0-7]{2}|[0-3][0-7]{2})|u[0-9a-fA-F]{4}))*"/
```
bd82: can we re-use the regexp from charLiteral instead of copy/paste?
Have a look at some MAKE_PATTERN examples in this file

Hawkurane: I was actually wondering if it was worth it to rename most of the regexp using the java specs and fragments/make_patterns to make it easier to read and translate directly from the java specs. But I am unsure if it has huge performance impacts.

We could make small PR's bit by bit to slowly use fragments

bd82: It is generally a good thing.
Note I implemented a better fragments composition system here:
- https://github.com/bd82/toml-tools/blob/master/packages/lexer/lib/tokens.js
- I think I was also trying to align with the TOML spec names for the lexer fragments there.

It may not be worth the effort to do so now, depends on how much time resources we have...
But we should at a minimum avoid new duplications...

Hawkurane: Yes I will try to do it then. Should we merge this one then do a later PR for fragments or should I just commit there?

bd82: I think that a big refactor for using more fragments should be done **separately**.

But that this PR for the stringLiteral + char patterns should avoid the **new** duplication.
So you should define a single fragment for re-use in char/stringliteral as part of this PR.

Basically do not introduce **more** technical debt in this PR
And consider resolving the existing technical debt at a later time depending on available time resources.

Hawkurane: I've managed to re-do the Character and String literals tokens. It was a bit tricky as MAKE_PATTERN transform \\ into \ so I had to go for quad-\'s ...

bd82: Yeah, the horrors of multiple levels of escaping 😠

packages/java-parser/src/tokens.js
```diff
@@ -39,7 +39,25 @@ FRAGMENT("HexDigit", "[0-9a-fA-F]");
...
+FRAGMENT("OctalDigit", "[0-7]");
+FRAGMENT("ZeroToThree", "[0-3]");
+FRAGMENT(
+  "OctalEscape",
+  "\\\\({{OctalDigit}}|{{ZeroToThree}}?{{OctalDigit}}{2})"
+);
+FRAGMENT("EscapeSequence", "\\\\[btnfr\"'\\\\]|{{OctalEscape}}");
+FRAGMENT(
+  "CharacterLiteral",
+  "'(?:[^\\\\']|(?:(?:{{EscapeSequence}})|\\\\u[0-9a-fA-F]{4}))'"
```
bd82: are we strongly aligning with the spec style here or not?
This does not use the spec terminology.
https://docs.oracle.com/javase/specs/jls/se11/html/jls-3.html#jls-CharacterLiteral

packages/java-parser/src/tokens.js
```diff
@@ -39,7 +39,25 @@ FRAGMENT("HexDigit", "[0-9a-fA-F]");
...
+);
+FRAGMENT(
+  "StringLiteral",
+  '"(?:[^\\\\"]|(?:(?:{{EscapeSequence}})|\\\\u[0-9a-fA-F]{4}))*"'
```
bd82: Not only does this not align with spec terminology
it also does duplication

packages/java-parser/src/tokens.js
```diff
@@ -199,12 +217,11 @@ createToken({
...
 createToken({
   name: "CharLiteral",
-  pattern: /'(?:[^\\']|\\(?:(?:[btnfr"'\\/]|[0-7]|[0-7]{2}|[0-3][0-7]{2})|u[0-9a-fA-F]{4}))'/
+  pattern: MAKE_PATTERN("{{CharacterLiteral}}")
 });
 createToken({
   name: "StringLiteral",
-  // TODO: align with the spec, the pattern below is incorrect
-  pattern: /"[^"\\]*(\\.[^"\\]*)*"/
+  pattern: MAKE_PATTERN("{{StringLiteral}}")
```
bd82: why do we need a stringLiteral fragment? can we not define that directly in the createToken?
as defined in: https://docs.oracle.com/javase/specs/jls/se11/html/jls-3.html#jls-StringLiteral

packages/java-parser/src/tokens.js
bd82: Do we not align with the spec because we use the [^] NOT group?

Hawkurane: I have reused the initial RegExp which was using the same [^] NOT group. I am trying to stay as close as possible to the spec while compromising parsing writing style with regexp writing style.

packages/java-parser/src/tokens.js
Hawkurane: We can, I was actually wondering whether to add this one FRAGMENT or not for aesthetic but let's go with your idea

packages/java-parser/src/tokens.js
Hawkurane: Could you please see the changes?

PR #214, opened 2019-06-13
Multiple line wrapping fixes

clement26695:


packages/prettier-plugin-java/src/printers/blocks-and-statements.js
```diff
@@ -17,24 +17,28 @@ const {
...
         ctx.RCurly[0]
-      ]);
+      );
     }
 
     if (
       hasTrailingComments(ctx.LCurly[0]) ||
       hasLeadingComments(ctx.RCurly[0])
     ) {
+      // TODO: to fix
```
Shaolans: A more detailed to do would be nice to easily understand what should be fixed.

packages/prettier-plugin-java/src/printers/blocks-and-statements.js
```diff
@@ -138,6 +142,7 @@ class BlocksAndStatementPrettierVisitor {
       ]);
     }
 
+    // TODO: fix line wrapping
```
Shaolans: same.

packages/prettier-plugin-java/src/printers/blocks-and-statements.js
```diff
@@ -161,6 +166,7 @@ class BlocksAndStatementPrettierVisitor {
     const expression = this.visit(ctx.expression);
     const switchBlock = this.visit(ctx.switchBlock);
 
+    // TODO: fix line wrapping
```
Shaolans: same

packages/prettier-plugin-java/src/printers/blocks-and-statements.js
```diff
@@ -213,6 +219,7 @@ class BlocksAndStatementPrettierVisitor {
         ? ""
         : " ";
 
+    // TODO: fix line wrapping
```
Shaolans: same

packages/prettier-plugin-java/src/printers/blocks-and-statements.js
```diff
@@ -261,6 +269,7 @@ class BlocksAndStatementPrettierVisitor {
...
```
Shaolans: same

packages/prettier-plugin-java/src/printers/blocks-and-statements.js
```diff
@@ -400,6 +409,7 @@ class BlocksAndStatementPrettierVisitor {
     const catchFormalParameter = this.visit(ctx.catchFormalParameter);
     const block = this.visit(ctx.block);
 
+    // TODO: fix line wrapping
```
Shaolans: same

packages/prettier-plugin-java/src/printers/prettier-builder.js
```diff
@@ -77,31 +77,44 @@ function processComments(docs) {
 }
 
 function concat(docs) {
-  return prettier.concat(processComments(docs));
+  const concatenation = prettier.concat(processComments(docs));
+  return concatenation.parts.length === 0 ? "" : concatenation;
```
Shaolans: IMO we should return ```undefined``` instead of ```""``` since we return an object and not a string.
It is also easier to use as we won't have to check with ```if(concat(docs) === "")``` but directory ```if(concat(docs))```

packages/prettier-plugin-java/src/printers/interfaces.js
```diff
@@ -68,22 +70,42 @@ class InterfacesPrettierVisitor {
...
+        ctx.LCurly[0],
+        ctx.RCurly[0]
+      );
+    }
+
+    if (
+      hasTrailingComments(ctx.LCurly[0]) ||
+      hasLeadingComments(ctx.RCurly[0])
+    ) {
+      // TODO: to fix
```
Shaolans: detail

packages/prettier-plugin-java/src/printers/classes.js
```diff
@@ -546,16 +529,27 @@ class ClassesPrettierVisitor {
...
+        ctx.LCurly[0],
+        ctx.RCurly[0]
+      );
+    }
+
+    if (
+      hasTrailingComments(ctx.LCurly[0]) ||
+      hasLeadingComments(ctx.RCurly[0])
+    ) {
+      // TODO: to fix
```
Shaolans: detail

packages/prettier-plugin-java/src/printers/prettier-builder.js
clement26695: I use the same "convention" as here :

https://github.com/clement26695/prettier-java/blob/35c11f4656af8ba1e591e52e25a6b0f42d95a683/packages/prettier-plugin-java/src/cst-printer.js#L69-L74

And IMO, it is the same for your second argument (testing if (elt) will check if elt === undefined/""/[] etc.)

But you're rigth, it migth be clearer...

PR #216, opened 2019-06-14
refactor: extract curly braces blocks logic into printer-utils

clement26695:
The logic to create a curly braces block was duplicated. This PR is to excract this logic to limit code duplication.

The putIntoCurlyBraces function is to print curly braces bloc with correct indentation, and handling empty curly braces block with just an inner comment.


bd82:
Very nice reduction of duplication 👍

PR #217, opened 2019-06-16
Fix comment only code

Shaolans:
The parser fails when given in input only commented code, this is due to the fact that we are attaching comments to token but since there is no token it breaks.
I tried a little "hack" by mocking a token where all comments will attach to it.

### DIFF
--- a/packages/java-parser/src/index.js
+++ b/packages/java-parser/src/index.js
@@ -22,31 +22,10 @@ function parse(inputText, entryPoint = "compilationUnit") {
     );
   }
 
-  const existToken = lexResult.tokens.length !== 0;
-  const tokenMock = [
-    {
-      image: "",
-      startOffset: 1,
-      endOffset: Infinity,
-      startLine: 1,
-      endLine: Infinity,
-      startColumn: 1,
-      endColumn: Infinity,
-      tokenTypeIdx: 91
-    }
-  ];
-
-  let cst;
-  if (existToken) {
-    parser.input = attachComments(lexResult.tokens, lexResult.groups.comments);
-    cst = parser[entryPoint]();
-  } else {
-    parser.input = tokenMock;
-    parser.input[0].trailingComments = [...lexResult.groups.comments];
-    cst = parser["emptyStatement"]();
-  }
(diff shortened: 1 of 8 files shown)


bd82:
Instead of creating a token mock you could try using an EOF token (End Of File).
to explicitly be consumed at the end of the root rule.

This Token type is returned by the parser when it is out of input
and is exported by Chevrotain.
- https://sap.github.io/chevrotain/documentation/4_8_0/globals.html#eof

Shaolans:
Thanks :+1:, I was not aware ot it

packages/java-parser/src/index.js
```diff
@@ -62,12 +41,26 @@ function parse(inputText, entryPoint = "compilationUnit") {
...
+    attachComments(lexResult.tokens, lexResult.groups.comments);
+  }
+
   return cst;
 }
 
 function attachComments(tokens, comments) {
   const attachComments = [...comments];
 
+  // edge case: when the file only contains comments
```
bd82: is that the only case where there is only a single token?
What if there is only a single empty statement ";" won't that also be the case of single token file?

```java
;
```

Or is that not valid in java?

packages/prettier-plugin-java/src/printers/packages-and-modules.js
```diff
@@ -14,7 +14,9 @@ const {
 
 class PackagesAndModulesPrettierVisitor {
   compilationUnit(ctx) {
-    return this.visitSingle(ctx);
+    const EOF = ctx.EOF[0];
+    delete ctx.EOF;
```
bd82: mutating the CTX does not make sense.
The next person who tries to visit the CST won't get the same results! 😢

Shaolans: You are totally right, I was getting an error of visitSingle since I added EOF. I could have just retrieve the children and call the visit on it.

packages/java-parser/src/index.js
Shaolans: Well the java compiler does not complains so it should be valid in Java I guess. I made changes to support single token.

PR #221, opened 2019-06-18
Prettier ignore support

### LINKED ISSUE #209: Handle prettierignore
We should handle prettier ignore when specified in comments (see https://prettier.io/docs/en/ignore.html)

Shaolans:
Fix #209

### DIFF
--- a/packages/java-parser/src/comments.js
+++ b/packages/java-parser/src/comments.js
@@ -1,4 +1,5 @@
 "use strict";
+const _ = require("lodash");
 
 function attachComments(tokens, comments) {
   const attachComments = [...comments];
@@ -144,13 +145,13 @@ function filterPrettierIgnore(comments) {
 }
 
 function shouldIgnore(node, comments, ignoredNodes) {
-  for (let i = 0; i < comments.length; i++) {
-    if (comments[i].extendedRange.endOffset === node.location.startOffset) {
-      ignoredNodes[comments[i].startOffset] = node;
-      return true;
-    }
+  const matchingComment = _.find(
+    comments,
+    comment => comment.extendedRange.endOffset === node.location.startOffset
+  );
+  if (matchingComment) {
+    ignoredNodes[matchingComment.startOffset] = node;
   }
-  return false;
 }
 
 function attachIgnoreNodes(ignoreComments, ignoredNodes) {
--- a/packages/java-parser/src/comments.js
+++ b/packages/java-parser/src/comments.js
@@ -83,6 +83,13 @@ function attachComments(tokens, comments) {
   return tokens;
 }
 
+/**
+ * Search where is the position of the comment in the token array by
+ * using dichotomic search.
+ * @param {*} tokens ordered array of tokens
+ * @param {*} comment comment token
+ * @return the position of the token next to the comment
+ */
 function findUpperBoundToken(tokens, comment) {
   let diff;
   let i;
@@ -104,6 +111,14 @@ function findUpperBoundToken(tokens, comment) {
   return i;
 }
 
+/**
+ * Extends each comments offsets to the left and the right in order to match the
+ * previous and next token offset. This allow to directly match the prettier-ignore
+ * comment to the correct CSTNode.
+ * @param {*} tokens ordered array of tokens
+ * @param {*} comments array of prettier-ignore comments
+ * @return prettier-ignore comment array with extended location
+ */
 function extendCommentRange(tokens, comments) {
   const ignoreComments = [...comments];
   let position;
--- a/packages/java-parser/src/comments.js
+++ b/packages/java-parser/src/comments.js
@@ -1,0 +1,156 @@
+"use strict";
+
+function attachComments(tokens, comments) {
+  const attachComments = [...comments];
+
+  // edge case: when the file contains only one token (;/if no token then EOF)
+  if (tokens.length === 1) {
+    attachComments.forEach(comment => {
+      if (comment.endOffset < tokens[0].startOffset) {
+        if (!tokens[0].leadingComments) {
+          tokens[0].leadingComments = [];
+        }
+        tokens[0].leadingComments.push(comment);
+      } else {
+        if (!tokens[0].trailingComments) {
+          tokens[0].trailingComments = [];
+        }
+        tokens[0].trailingComments.push(comment);
+      }
+    });
+    return tokens;
+  }
+
+  // edge case: when the file start with comments, it attaches as leadingComments to the first token
+  const firstToken = tokens[0];
+  const headComments = [];
+  while (
+    attachComments.length > 0 &&
+    attachComments[0].endOffset < firstToken.startOffset
+  ) {
+    headComments.push(attachComments[0]);
+    attachComments.splice(0, 1);
+  }
+
+  if (headComments.length > 0) {
+    firstToken.leadingComments = headComments;
+  }
+
+  // edge case: when the file end with comments, it attaches as trailingComments to the last token
+  const lastToken = tokens[tokens.length - 1];
+  const tailComments = [];
+  while (
+    attachComments.length > 0 &&
+    attachComments[attachComments.length - 1].startOffset > lastToken.endOffset
+  ) {
+    tailComments.push(attachComments[attachComments.length - 1]);
+    attachComments.splice(attachComments.length - 1, 1);
+  }
+
+  if (tailComments.length > 0) {
+    lastToken.trailingComments = tailComments.reverse();
+  }
+
+  let currentToken = 0;
+  attachComments.forEach(element => {
+    // find the correct position to place the comment
+    while (
+      !(
+        element.startOffset > tokens[currentToken].endOffset &&
+        element.endOffset < tokens[currentToken + 1].startOffset
+      )
+    ) {
+      currentToken++;
+    }
+
+    // attach comment to the closest token
+    if (
+      element.startOffset - tokens[currentToken].endOffset <
+      tokens[currentToken + 1].startOffset - element.endOffset
+    ) {
+      if (!tokens[currentToken].hasOwnProperty("trailingComments")) {
+        tokens[currentToken].trailingComments = [];
+      }
+      tokens[currentToken].trailingComments.push(element);
+    } else {
+      if (!tokens[currentToken + 1].hasOwnProperty("leadingComments")) {
+        tokens[currentToken + 1].leadingComments = [];
+      }
+      tokens[currentToken + 1].leadingComments.push(element);
+    }
+  });
+
+  return tokens;
+}
+
+function findUpperBoundToken(tokens, comment) {
+  let diff;
+  let i;
+  let current;
+
+  let len = tokens.length;
+  i = 0;
+
+  while (len) {
+    diff = len >>> 1;
+    current = i + diff;
+    if (tokens[current].startOffset > comment.startOffset) {
+      len = diff;
+    } else {
+      i = current + 1;
+      len -= diff + 1;
+    }
+  }
+  return i;
+}
+
+function extendCommentRange(tokens, comments) {
+  const ignoreComments = [...comments];
+  let position;
+  ignoreComments.forEach(comment => {
+    position = findUpperBoundToken(tokens, comment);
+    comment.extendedRange = {};
+    comment.extendedRange.startOffset =
+      position - 1 < 0 ? comment.startOffset : tokens[position - 1].endOffset;
+    comment.extendedRange.endOffset =
+      position == tokens.length
+        ? comment.endOffset
+        : tokens[position].startOffset;
+  });
+  return ignoreComments;
+}
+
+function filterPrettierIgnore(comments) {
+  return [...comments].filter(comment =>
+    comment.image.match(/\/\/(\s*)prettier-ignore(\s*)/gm)
+  );
+}
+
+function shouldIgnore(node, comments, ignoredNodes) {
+  for (let i = 0; i < comments.length; i++) {
+    if (comments[i].extendedRange.endOffset === node.location.startOffset) {
+      ignoredNodes[comments[i].startOffset] = node;
+      return true;
+    }
+  }
+  return false;
+}
+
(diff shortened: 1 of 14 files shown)


packages/prettier-plugin-java/src/cst-printer.js
```diff
@@ -73,6 +77,17 @@ class CstPrettierPrinter extends BaseJavaCstVisitor {
         return "";
       }
 
+      if (ctx.ignore) {
+        try {
+          return buildOriginalText(retrieveNodesToken(ctx));
+        } catch (e) {
+          // eslint-disable-next-line no-console
+          console.error(
+            "There might be a problem with prettier-ignore, check cst-printer.js"
```
clement26695: Maybe change the message for end-users.

clement26695:
LGTM. Does it have a big impact on performance?

Shaolans:
> LGTM. Does it have a big impact on performance?
Almost none when there is no // prettier-ignore comment and I am still pretty confident that the impact is low since the use of prettier-ignore is scarce.

@bd82 Do you have anytime to review it ? Since we want to release soon a new version We'd like this feature to be integrated for the release.

packages/prettier-plugin-java/src/cst-printer.js
Shaolans: This should not be happening, if an error is thrown this is probably due to a forgotten case or e.g.

clement26695: Yes, so maybe put something like: "Please report an issue on..."

Shaolans: Yes, will do :+1:

packages/prettier-plugin-java/src/printers/printer-utils.js
```diff
@@ -443,6 +443,67 @@ function isShiftOperator(tokens, index) {
...
+    ctx.children[child].forEach(subctx => {
+      tokens.push(...retrieveNodesTokenRec(subctx));
+    });
+  });
+  return tokens;
+}
+
+function buildOriginalText(tokens) {
+  if (tokens.length == 0) {
+    ("");
```
bd82: is a return missing in the first branch?

packages/prettier-plugin-java/src/printers/printer-utils.js
```diff
@@ -443,6 +443,67 @@ function isShiftOperator(tokens, index) {
...
+      originalText.push(hardline);
+      startLine--;
+    }
+    if (tokens[i].startLine - lastToken.startLine > 0) {
+      space = tokens[i].startColumn - 1;
+    } else {
+      space = tokens[i].startColumn - lastToken.endColumn - 1;
+    }
+    while (space > 0) {
+      originalText.push(" ");
```
bd82: does not this assume that each space is a regular space?
What if the space was a tab?

packages/prettier-plugin-java/src/printers/printer-utils.js
```diff
@@ -443,6 +443,67 @@ function isShiftOperator(tokens, index) {
...
+  }
+  Object.keys(ctx.children).forEach(child => {
+    ctx.children[child].forEach(subctx => {
+      tokens.push(...retrieveNodesTokenRec(subctx));
+    });
+  });
+  return tokens;
+}
+
+function buildOriginalText(tokens) {
```
bd82: Building the original text may be error prone.
Could we get a substring of the original text given start/end offsets of the CSTNode
and slice it into the new "formatted" output we are creating instead?

packages/java-parser/src/comments.js
```diff
@@ -0,0 +1,156 @@
...
+}
+
+function filterPrettierIgnore(comments) {
+  return [...comments].filter(comment =>
+    comment.image.match(/\/\/(\s*)prettier-ignore(\s*)/gm)
+  );
+}
+
+function shouldIgnore(node, comments, ignoredNodes) {
+  for (let i = 0; i < comments.length; i++) {
```
bd82: are we using imperative logic for performance reasons?

packages/java-parser/src/comments.js
```diff
@@ -0,0 +1,156 @@
...
+      position == tokens.length
+        ? comment.endOffset
+        : tokens[position].startOffset;
+  });
+  return ignoreComments;
+}
+
+function filterPrettierIgnore(comments) {
+  return [...comments].filter(comment =>
+    comment.image.match(/\/\/(\s*)prettier-ignore(\s*)/gm)
```
bd82: can't a prettier-ignore comment be a multi line comment too? (/* prettier ignore */)

packages/java-parser/src/comments.js
```diff
@@ -0,0 +1,156 @@
...
+      len = diff;
+    } else {
+      i = current + 1;
+      len -= diff + 1;
+    }
+  }
+  return i;
+}
+
+function extendCommentRange(tokens, comments) {
```
bd82: this needs docs i dont understand what is happening here.

packages/java-parser/src/comments.js
```diff
@@ -0,0 +1,156 @@
...
+        tokens[currentToken + 1].leadingComments = [];
+      }
+      tokens[currentToken + 1].leadingComments.push(element);
+    }
+  });
+
+  return tokens;
+}
+
+function findUpperBoundToken(tokens, comment) {
```
bd82: needs comments/docs i do not understand what is the intent

packages/java-parser/src/comments.js
Shaolans: yes you are right, I forgot that case :+1:

packages/java-parser/src/comments.js
Shaolans: yes, as it is not necessary to continue further if the comment is already matched with the node.

packages/java-parser/src/comments.js
Shaolans: The offsets on nodes are defined by the startOffset of the first token and the endOffset of the last token. However between a token and another one (comment in our case) there might be some spaces/tabs.
Thus, by extending the "comments" range to the endOffset of the previous token and the startOffset of the next token. We are able to know to which node we should attach as IMO would be much harder without it and will add more computing power if we do it when we traverse all the cst.

Shaolans: I will comment and add docs on it. But I already planned to document the project.

packages/java-parser/src/comments.js
Shaolans: It returns the indice of where the comment is located between the tokens with a logarithmic complexity.
Since the token array is sorted, we can just run a dichotomic search.

packages/prettier-plugin-java/src/printers/printer-utils.js
Shaolans: Oh right I have not thought about it since my configured my IDE to consider tab as spaces.

packages/prettier-plugin-java/src/printers/printer-utils.js
Shaolans: It could work but the start/end offsets on CSTNode only consider the non comment token so if we have 
```java
// prettier-ignore
int a = 1; // my int
```
it will slice and give
```java
int a = 1;
```
Thus removing the comments.

Shaolans: A workaround would be to retrieve all the tokens on the nodes and check if there is a leading/trailing comment. WDYT ?

bd82: > A workaround would be to retrieve all the tokens on the nodes and check if there is a leading/trailing comment. WDYT ?

In general this still seems less error prone.
- Why do we need to check all Tokens on Node though? won't the first and last tokens be enough?

packages/java-parser/src/comments.js
bd82: I meant why are we using for loops instead of forEach and high order functions?

Shaolans: If we use a forEach it will still iterate through all elements even after the node is matched.

packages/prettier-plugin-java/src/printers/printer-utils.js
Shaolans: Yes it will, this is what I am doing right now https://github.com/jhipster/prettier-java/pull/221/files#diff-21eac804fc21eb2d2bb5849e0cf5e7eeR474. I check the first and last token.

packages/java-parser/src/comments.js
bd82: Lodash's forEach can exit early: https://lodash.com/docs/4.17.11#forEach
But that is besides the point. it looks like you are trying to **find** an element in a collection and do something with that element.
- https://lodash.com/docs/4.17.11#find may be a relevant tool in this use case.

So you have to consider, do you go with the more declarative/functional style
as it is less error prone and less code, or do you go with the imperative style with regular loops and such.

Each has its pros and cons, in I personally tend to only use regular imperative loops in performance critical code, otherwise I prefer to pay a small overhead but gain higher quality source code (imho).

packages/prettier-plugin-java/src/printers/printer-utils.js
bd82: sounds good!

packages/prettier-plugin-java/src/printers/printer-utils.js
bd82: if we chop of the relevant part from the original string this issue would be resolved.

Shaolans: Yup ! I implemented it and it is working pretty well 👍

packages/java-parser/src/comments.js
Shaolans: Interesting, I will keep it in my mind. 👍 
I will probably do as you suggest it.
Do you have any other remark on this PR ? If not I will merge it after changing this piece of code.

packages/java-parser/src/comments.js
bd82: Cool, please evaluate if you can implement something similar by using utilities such as:
- https://lodash.com/docs#sortedIndexBy

packages/java-parser/src/comments.js
bd82: O.K I think I understand. You want to know the "real" token immediately following a comment and the "real" token just before a comment.

Would constructing a singular (sorted) combined Token Vector that combines both the "real" tokens and the comment tokens assist in this regard and/or be simpler?

Also does the logic below handle the edge case where there is a sequence of comment tokens?

packages/java-parser/src/index.js
```diff
@@ -24,6 +29,13 @@ function parse(inputText, entryPoint = "compilationUnit") {
 
   parser.input = lexResult.tokens;
 
+  // prettier-ignore support
```
bd82: I think you accidentally disabled formatting on this next line 😄

packages/prettier-plugin-java/src/printers/printer-utils.js
```diff
@@ -443,6 +443,52 @@ function isShiftOperator(tokens, index) {
   return "none";
 }
 
+function retrieveNodesToken(ctx) {
```
bd82: This recursive re-collecting of all the tokens happens for **every** ignore-comment?
This does not seem to be very effective, but maybe it is o.k. because there should be very few ignore comments in most input sources...

I also think you only need the first and last tokens from a CSTNode (specifically their start/end offsets) so perhaps you do not really have to collect all the tokens just the min/max offsets?


But do not we already have the CSTNodes start/end offsets in the new location property? so perhaps this logic is redundant?

Shaolans: > This recursive re-collecting of all the tokens happens for every ignore-comment?
> This does not seem to be very effective, but maybe it is o.k. because there should be very few ignore comments in most input sources...

I am using this function to retrieve all the tokens as we might need it for other functionnality. I am more thinking about organising imports. We will need this kind of function as we want to be sorted by the name of it.

> I also think you only need the first and last tokens from a CSTNode (specifically their start/end offsets) so perhaps you do not really have to collect all the tokens just the min/max offsets?

I agree, I was just using the function already written I can make another one to just retrieve the first and last token.

> But do not we already have the CSTNodes start/end offsets in the new location property? so perhaps this logic is redundant?

We do have the location but as I said on another conversation, the location on nodes does not include comments.
For instance: 
```java
// prettier-ignore
int a; // my int
```
The location on the CSTNode associated to ```int a;```` will start on ```i``` and end on the semicolon ```;``` and won't cover ```// my int``` so if I cut it won't cut the comment.

packages/java-parser/src/index.js
Shaolans: woops !

packages/prettier-plugin-java/src/printers/printer-utils.js
bd82: O.k. I understand why you are doing this function now, just be wary about re-creating an augmented token vector with contents you do not actually need.

It is is highly inefficient...

Shaolans: I am now using a function that avoid to rebuild a token vector.

packages/prettier-plugin-java/src/cst-printer.js
```diff
@@ -79,7 +79,17 @@ class CstPrettierPrinter extends BaseJavaCstVisitor {
 
       if (ctx.ignore) {
         try {
-          return buildOriginalText(retrieveNodesToken(ctx), this.originalText);
+          return buildOriginalText(
+            getCSTNodeToken(
```
bd82: if this is an "heavy" process why do we call it twice?
clone and sort externally from the "heavy" collection 😄

bd82: Ah I see it returns only a single token, but iterates over the structure twice.
I would try to make it return two tokens (first and last) in one call and avoid the multiple iterations.

Shaolans: I was trying to make a "generic" function bu it seems that we should not in this case.
Corrected as you suggested.

Issue #218, opened 2019-06-18
format differences with PrettierJS


Hawkurane:
We want to stay as close as possible to the reformatting style of Prettier JS

Below are some examples that should be reviewed:
- TODO
```
var databaseSizeBeforeCreate = userRepository.findAll().collectList().block().size();
// PrettierJS =>
var databaseSizeBeforeCreate = userRepository
  .findAll()
  .collectList()
  .block()
  .size();
// Prettier-Java =>
var databaseSizeBeforeCreate = userRepository.findAll()
  .collectList()
  .block()
  .size();
```
- TODO
```
var LONGER_LONGER_LONGER_LONGER_LONGER_LOCALE_REQUEST_ATTRIBUTE_NAME = AngularCookieLocaleContextResolver.class.getName() + ".LOCALE";
// PrettierJS =>
var LONGER_LONGER_LONGER_LONGER_LONGER_LOCALE_REQUEST_ATTRIBUTE_NAME =
  AngularCookieLocaleContextResolver.class.getName() + ".LOCALE";

private static final String LOCALE_REQUEST_ATTRIBUTE_NAME = AngularCookieLocaleContextResolver.class.getName()+ ".LOCALE";
// Prettier-Java =>
private static final String LOCALE_REQUEST_ATTRIBUTE_NAME = AngularCookieLocaleContextResolver
  .class
    .getName()
    + ".LOCALE";

// It seems prettier tries to put the second variable on a second line and then only try to break the second variable if it still exceeds the 80 character line
```

- PR #223 (merged)

```
if(aVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongName && anotherVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongName)
{ // do something
}
// PrettierJS =>
if (
  aVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongName &&
  anotherVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongName
) {
  // do something
}
// Prettier-Java =>
if (aVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongName
  && anotherVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongName) { // do something
}
```

- PR #222 :

clement26695:
I put it here, as I think the change has to be done in the same place. We have also this bug :

```java
public static void main(String[] args) {
[...]

PR #222, opened 2019-06-18
WIP: little tweak, softline in variable initialization

Hawkurane:
This adds a soft line when initializing a variable. This is get closer to PrettierJS behaviour. See #218.

### DIFF
--- a/packages/prettier-plugin-java/src/printers/expressions.js
+++ b/packages/prettier-plugin-java/src/printers/expressions.js
@@ -233,7 +233,7 @@ class ExpressionsPrettierVisitor {
       : [];
     return rejectAndConcat([
       rejectAndConcat(unaryPrefixOperator),
-      primary,
+      primary, //System.out.println("...");  //userRepository.findAll().collectList().block().size();
       rejectAndConcat(unarySuffixOperator)
     ]);
   }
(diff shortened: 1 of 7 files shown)


Hawkurane:
Currently, the java printer is not able to break a variable declaration into two lines such as:
```Java
private static int some_variable = 
    some_random_value;
```
I've tried to use a few tricks to make it doable but I can't seem to be able to handle the priority of grouping.

```Java
// Should be
private static int some_variable = some_random_function(
    value1,
    value2
);

// This breaks the line when we are using a binary operator instead of breaking the argument list.
private static int some_variable =
    some_random_function(value1, value2) + "literal_value";

// What it does no matter what
private static int some_variable =
    some_random_function(value1, value2);
```

This suits PrettierJS in this specific case but overrides many other cases (see 2nd example)

clement26695:
The issue is that you cannot tell which group should break in priority with the prettier API : see https://github.com/prettier/prettier/blob/master/commands.md#group

> Groups are usually nested, and the printer will try to fit everything on one line, but if it doesn't fit it will break the outermost group first and try again. It will continue breaking groups until everything fits (or there are no more groups to break).

Hawkurane:
Indeed it looks like the outermost will always break in priority.
In our case it seems to vary, sometimes the innermost breaks first...

packages/prettier-plugin-java/src/printers/expressions.js
```diff
@@ -277,17 +278,25 @@ class ExpressionsPrettierVisitor {
       if (ctx.primaryPrefix[0].children.This !== undefined) {
         firstSeparator = "";
       }
+      // console.log("=====================");
```
bd82: lets not merge commented out code 😄

Hawkurane: Yes it is still WIP because I am still trying to fix 2 bugs and I still need the printing.

Hawkurane:
I am closing this PR as this is currently too difficult. I will rework on this at a later date with smaller PR's because this one is dependent of others

PR #223, opened 2019-06-20
Fix expression print

Shaolans:
Fix the breaking expression discussed [here](https://github.com/jhipster/prettier-java/issues/218)

Example:
input:
```java
public class Args {

    if(test == 1 && test == 1 && test == 1 && test == 1 && test == 1 && test == 1 && test == 1 && test == 1) {

    }
}
}
```

output:
```java
public class Args {

    if (
      test == 1 &&
      test == 1 &&
      test == 1 &&
      test == 1 &&
      test == 1 &&
      test == 1 &&
      test == 1 &&
      test == 1
    ) {}
  }
}

```

~~However, it will only breaks on && and ||.
if we have ```if(aVeryVeryVeryVeryVeryVeryVeryVeryVeryLongVariable == aVeryVeryVeryVeryVeryVeryVeryVeryVeryLongVariable2)```, it won't break on ```==```.
Is it bothersome ? I am not sure... It will be hard to tell how we should break each operator since we just have an array on all operators on the cstnode.~~
I am working with @clement26695 to make it work :+1:

### DIFF
--- a/packages/prettier-plugin-java/src/printers/expressions.js
+++ b/packages/prettier-plugin-java/src/printers/expressions.js
@@ -213,8 +213,6 @@ class ExpressionsPrettierVisitor {
         let separator = " ";
         if (token.image === "&&" || token.image === "||") {
           separator = line;
-        } else {
-          separator = " ";
         }
         segment.push(
           rejectAndConcat([
--- a/packages/prettier-plugin-java/src/printers/expressions.js
+++ b/packages/prettier-plugin-java/src/printers/expressions.js
@@ -226,7 +226,7 @@ class ExpressionsPrettierVisitor {
         );
       }
     }
-    return group(rejectAndJoin(" ", segment));
+    return group(concat([softline, rejectAndJoin(" ", segment), softline]));
   }
 
   unaryExpression(ctx) {
--- a/packages/prettier-plugin-java/src/printers/expressions.js
+++ b/packages/prettier-plugin-java/src/printers/expressions.js
@@ -18,7 +18,7 @@ const {
   findDeepElementInPartsArray,
   isExplicitLambdaParameter,
   putIntoBraces,
-  sortGroupTokens,
+  separateTokensIntoGroups,
   isShiftOperator
 } = require("./printer-utils");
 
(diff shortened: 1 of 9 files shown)


packages/prettier-plugin-java/test/unit-test/binary_expressions/_output.java
```diff
@@ -2,8 +2,7 @@
 
   public void binaryOperationThatShouldBreak() {
     System.out.println(
-      "This operation with two very long string should break"
-        + "in a very nice way"
+      "This operation with two very long string should break" + "in a very nice way"
```
clement26695: If it fits in one line, the string should be updated

packages/prettier-plugin-java/src/printers/expressions.js
```diff
@@ -210,13 +210,17 @@ class ExpressionsPrettierVisitor {
         }
         i++;
       } else if (matchCategory(token, "'BinaryOperator'")) {
+        let separator = " ";
+        if (token.image === "&&" || token.image === "||") {
```
clement26695: I think it shoud include `|`, `&` and `^` tokens

packages/prettier-plugin-java/src/printers/expressions.js
```diff
@@ -210,13 +210,17 @@ class ExpressionsPrettierVisitor {
...
+          separator = line;
+        } else {
+          separator = " ";
```
clement26695: No need to assign the same value as before.

clement26695:
Shouldn't the catch Clause be updated too?

clement26695:
Maybe also add some test case which should not break? Like:

```java
if (test1 == true && test2 != null)
```

Shaolans:
> Shouldn't the catch Clause be updated too?

This is already handled by a different "rule", but it add the ```|``` on the new line and not on the end of line. However when I add it to the end of line, my visual studio linter for Java does not recognize the "Exception"
example
```java
try {} catch (
      Exception
      | Exception
      | Exception
      | Exception
      | Exception
      | Exception
      | Exception
      | Exception
      | Exception
      | Exception e
    ) {}
```

```java
try {} catch (
      Exception |
      Exception |
      Exception |
      Exception |
      Exception |
      Exception |
      Exception |
      Exception |
      Exception |
      Exception e
    ) {}
```

Shaolans:
this looks good now.

Example:
input:
```java
if(test == 1 && test == 1 && test == 1 && test == 1 && test == 1 && test == 1 && test == 1 && test == 1 && test == 1 && mljkejkoerjkotjerjtlejlrtj == elkrjtklerjtklejklrtjkeljtklejklrtjklejrtkljeklrjtklejrkltjelkrtjl) {

    }
```
output:
```java
if (
      test == 1 &&
      test == 1 &&
      test == 1 &&
      test == 1 &&
      test == 1 &&
      test == 1 &&
      test == 1 &&
      test == 1 &&
      test == 1 &&
      mljkejkoerjkotjerjtlejlrtj ==
        elkrjtklerjtklejklrtjkeljtklejklrtjklejrtkljeklrjtklejrkltjelkrtjl
    ) {}
```

packages/prettier-plugin-java/src/printers/expressions.js
```diff
@@ -153,74 +154,63 @@ class ExpressionsPrettierVisitor {
...
-        if (
-          sortedTokens[i + 2] &&
-          sortedTokens[i + 2].image === ">" &&
-          sortedTokens[i + 1].startOffset ===
-            sortedTokens[i + 2].startOffset - 1
+
+    const { groups, sortedBinOps } = separateTokensIntoGroups(ctx);
+    const allSegment = [];
+    let tmpSegment = [];
+
```
clement26695: I would try to clarify what those variables mean. 

Maybe:
- groupsOfOperator instead of groups, 
- sortedBinaryOperators instead of sortedBinOps,
- segmentsSplittedByBinaryOperator instead of allSegments
- currentSegment instead of tmpSegment

WDYT?

Shaolans: yep those variables fits it better :+1:

PR #228, opened 2019-06-25
grouped annotation in a single line when local variable is defined

### LINKED ISSUE #227: Annotation on variable declaration
Right now, the current format for Annotations with Local variables is as such:

```Java
@SuppressWarnings(
  "unchecked"
) V[][] tmpArray = (V[][]) new Object[rowList.size()][columnList.size()];
```

While I think it should be like this:
```Java
@SuppressWarnings("unchecked")
V[][] tmpArray = (V[][]) new Object[rowList.size()][columnList.size()];
```

Hawkurane:
Fixes #227

### DIFF
--- a/packages/prettier-plugin-java/src/printers/blocks-and-statements.js
+++ b/packages/prettier-plugin-java/src/printers/blocks-and-statements.js
@@ -61,9 +61,7 @@ class BlocksAndStatementPrettierVisitor {
       group(
         concat([
           variableModifiers.length > 0 ? softline : [],
-          localVariableType,
-          " ",
-          variableDeclaratorList
+          rejectAndJoin(" ", [localVariableType, variableDeclaratorList])
         ])
       )
     ]);
--- a/packages/prettier-plugin-java/src/printers/blocks-and-statements.js
+++ b/packages/prettier-plugin-java/src/printers/blocks-and-statements.js
@@ -56,15 +56,12 @@ class BlocksAndStatementPrettierVisitor {
     const variableModifiers = this.mapVisit(ctx.variableModifier);
     const localVariableType = this.visit(ctx.localVariableType);
     const variableDeclaratorList = this.visit(ctx.variableDeclaratorList);
-    return rejectAndJoin(" ", [
-      rejectAndJoin(" ", variableModifiers),
-      group(
-        concat([
-          variableModifiers.length > 0 ? softline : [],
-          rejectAndJoin(" ", [localVariableType, variableDeclaratorList])
-        ])
-      )
-    ]);
+    return group(
+      rejectAndJoin(line, [
+        rejectAndJoin(" ", variableModifiers),
+        rejectAndJoin(" ", [localVariableType, variableDeclaratorList])
+      ])
+    );
   }
 
   localVariableType(ctx) {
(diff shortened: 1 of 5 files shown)


packages/prettier-plugin-java/src/printers/blocks-and-statements.js
```diff
@@ -49,18 +49,23 @@ class BlocksAndStatementPrettierVisitor {
 
   localVariableDeclarationStatement(ctx) {
     const localVariableDeclaration = this.visit(ctx.localVariableDeclaration);
-    return rejectAndConcat([localVariableDeclaration, ctx.Semicolon[0]]);
+    return group(rejectAndConcat([localVariableDeclaration, ctx.Semicolon[0]]));
```
clement26695: I don't think there is the need for a group here, as there are no line between the localVariableDeclaration and the semicolon.

packages/prettier-plugin-java/src/printers/blocks-and-statements.js
```diff
@@ -49,18 +49,23 @@ class BlocksAndStatementPrettierVisitor {
...
     const localVariableType = this.visit(ctx.localVariableType);
     const variableDeclaratorList = this.visit(ctx.variableDeclaratorList);
-
     return rejectAndJoin(" ", [
       rejectAndJoin(" ", variableModifiers),
-      localVariableType,
-      variableDeclaratorList
+      group(
+        concat([
+          variableModifiers.length > 0 ? softline : [],
```
clement26695: As rejectAndJoin will get rid of the variableModifiers if there are none, you can just use a line at line 59 instead of a " " separator, and group at an upper level.

packages/prettier-plugin-java/src/printers/blocks-and-statements.js
```diff
@@ -49,18 +49,23 @@ class BlocksAndStatementPrettierVisitor {
...
+          localVariableType,
+          " ",
```
clement26695: You can then use rejectAndJoin here

packages/prettier-plugin-java/src/printers/blocks-and-statements.js
Hawkurane: Fixed, thank you it seems I forgot to remove it.

packages/prettier-plugin-java/src/printers/blocks-and-statements.js
Hawkurane: Currently, using annotation in a field declaration and in a local variable declaration have different bevhaviour:
As a field:
```Java
@RandomBreakingAnnotation(
  one = "One",
  two = "Two",
  three = "Three",
  four = "Four",
  five = "Five"
)
Integer[][] annotedArray = (Integer[][]) new Object[4][2];
```

as a local variable:
```Java
@RandomBreakingAnnotation(
  one = "One",
  two = "Two",
  three = "Three",
  four = "Four",
  five = "Five"
) Integer[][] annotedArray = (Integer[][]) new Object[4][2];
```

By changing the " " into a line and removing `variableModifiers.length > 0 ? softline : []`, we get thet first behaviour all the time. I think it is better to have an uniformized formatting but I think the second behaviour is prettier but that's my own opinion. "Fixing" the field declaration to make it look like the local variable declaration one will need another PR. WDYT

clement26695: I personally prefer the first one 😄

Hawkurane: Alright, I've gone along your review !

Issue #231, opened 2019-06-27
Multiline Comments get reduced to one line


PiiSmith:
In a block of more than one line of comments using `//`, only the last line is preserved. All the other lines get deleted.

Example:

When you start with this comment block:
```
//     Some comment
//     Some comment
//     Some comment
//     Some comment
//     Last comment
```

It gets reduced to this:
```
//     Last comment
```

PiiSmith:
Closed because the wrong version (npm and not locally built from master) was used.

Issue #191, opened 2019-05-07
Releasing new versions


bd82:
I will update the contribution guide with relevant instructions and someone else will hopefully trigger a release of the new versions.

bd82:
I am simply not getting around to this...

## Some hints to whomever wants to handle this.
The release process should very similar to the one I use in Chevrotain.
- https://github.com/SAP/chevrotain/blob/master/CONTRIBUTING.md#release-process

The build process would need to be modified to have a special release build that would be trigger on tag pushes, This is how I do it with CircleCI, I am sure something similar is possible in AzureDevOps
- [Release Build Definition](https://github.com/SAP/chevrotain/blob/94a21a767d50378bedf8404a7a049d02e7bb936c/.circleci/config.yml#L33-L49)
- [Release Build Trigger](https://github.com/SAP/chevrotain/blob/94a21a767d50378bedf8404a7a049d02e7bb936c/.circleci/config.yml#L58-L63)

Additionally for the release build trigger to work an NPM_TOKEN env variable must be made
available during the build. It is possible to obtain one from npmjs.com web UI once logged in.

jdubois:
Yes I believe it is time for a new release, this is what will make the project go final.

I will have a look today but I'm probably not the best one - who would be interested? I think this is worth some bug bounty, I'll configure this later once I've seen the amount of work needed.

Shaolans:
@bd82 Hello, I think we should release now a new version of prettier-java. WDYT ?
I think we have currently have a pretty good result if not perfect and it would be great to have some user feedbacks. Plus, I am seeing people trying to use but cannot since the npm prettier-java is the old one.

clement26695:
You can look at clement26695/jhipster-sample-app#2 to look at the result of formatting on the JHipster sample app

clement26695:
@Shaolans, I think we should revert b602ad80d0d94e09276a9e8cac2e2bb3fd919482 first

bd82:
@clement26695 @Shaolans 

I would prioritize sorting out an automated release process.
Then each release would be a simple command and you can release as many versions as you want per day...

bd82:
I just released 0.3.1

- The release command is : "yarn run lerna:version" (you can see it in the package.json).
- The trigger happen in travis and should be migrated to AzureDevops
- The auth uses npm_tokens (please add one of your own on azureDevops so I can remove mine).
- The release command is not enough, a process should be built around it (e.g updating changelog or even generating a changelog automatically).

[...]

PR #233, opened 2019-06-27
add a warning about the outdated npm package

n-peugnet:
As the NPM package is outdated and a lot of people, including myself, are trying to use the plugin through NPM, I suggest to add a warning at the top of the prettier-plugin-java's README.

Probably fixing #231

### DIFF
--- a/packages/prettier-plugin-java/README.md
+++ b/packages/prettier-plugin-java/README.md
@@ -1,3 +1,10 @@
+<p align="center">
+    :construction: Work in Progress! :construction:
+</p>
+<p align="center">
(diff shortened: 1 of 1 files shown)


bd82:
Warnings are good, but I really think #191  should have been implemented already and releases should be made on a regular basis... Maybe I will have some time to assist with this on the weekend...

bd82:
Can you try 0.3.1, I quickly released it even though there is no proper release procedure yet.
Does it work properly?

n-peugnet:
I can try when I get home but I'm currently stuck in public transportations

n-peugnet:
So it is working correctly and not outdated anymore so I am closing this PR

PR #249, opened 2019-08-28
Delete 'fs' import because that is not used

maxfie1d:
Our project that uses `java-parser` was failed to build.
And we found `java-parser` imports unused `fs` library, which was the cause of the problem.


clement26695:
Thank you for your contribution!

PR #250, opened 2019-09-05
Fix multi-lines comments

clement26695:
Fixes multi-lines comments as:

```
public class Test<TEST, RZE>
    extends Go
    implements Alpha, Beta, Alpha, Alpha, Alpha, Alpha, Alpha, Alpha {

    /*
    comment
     */
    @Pointcut("within(@org.springframework.stereotype.Repository *)" +
        " || within(@org.springframework.stereotype.Service *)" +
        " || within(@org.springframework.web.bind.annotation.RestController *)")
    public void springBeanPointcut() {
        // Method is empty as this is just a Pointcut, the implementations are in the advices.
    }
}
```

The annotation was previously printed on one line. 

It was the last "major" bug I had in mind

### DIFF
--- a/packages/prettier-plugin-java/src/printers/blocks-and-statements.js
+++ b/packages/prettier-plugin-java/src/printers/blocks-and-statements.js
@@ -17,7 +17,8 @@ const {
   getBlankLinesSeparator,
   rejectSeparators,
   putIntoBraces,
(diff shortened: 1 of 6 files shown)


clement26695:
I don't find any use case where it is really relevant

PR #255, opened 2019-09-10
Fix/binary operation indent

### LINKED ISSUE #251: Some binary expression should be indented when they break
Example:

```java
public class DateTimeWrapper implements Serializable {

    public boolean equals(Object o) {
        return !(dateTimeWrapper.getId() == null || getId() == null) &&
            Objects.equals(getId(), dateTimeWrapper.getId());
    }
```

is printed as:
```java
public class DateTimeWrapper implements Serializable {

    public boolean equals(Object o) {
        return !(dateTimeWrapper.getId() == null || getId() == null) &&
        Objects.equals(getId(), dateTimeWrapper.getId());
    }
```

clement26695:
Should fix #251 and #218 and #261

### DIFF
--- a/packages/prettier-plugin-java/src/cst-printer.js
+++ b/packages/prettier-plugin-java/src/cst-printer.js
@@ -32,13 +32,13 @@ class CstPrettierPrinter extends BaseJavaCstVisitor {
     // TODO: this methods should be defined on the prototype
     // defining as instance members **after** the validations to avoid
     // false positive errors on redundant methods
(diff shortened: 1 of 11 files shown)


Shaolans:
@clement26695  Are you using a second parameter for the visit method in order to give a context ?

clement26695:
yes

clement26695:
There is an issue with this PR, as some formatting is not stable. I'll have to take a look

PR #256, opened 2019-09-11
update pipeline to automate release

clement26695:
Ping @pascalgrimaud 

I tried to make the release process easier. Please let me know if we can do something better :)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,7 +21,7 @@ jobs:
 - job: Test
   strategy:
     matrix:
-      node_12: 
+      node_12:
         node_version: '12.8.1'
       node_10:
         node_version: '10.16.3'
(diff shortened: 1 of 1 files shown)


pascalgrimaud:
Did you test this on a sample project, with your own NPM_TOKEN, on your own repo ?

clement26695:
Yes, I tested it, but not the complete workflow (I tested separately the correct triggering of the job, and that the packages were correctly published). I tested the complete workflow here: https://dev.azure.com/clementdessoude/Test/_build/results?buildId=380&view=results.

The packages @cdessoude/test-parser and @cdessoude/test-printer should be updated on the npm registry with versions 0.4.0.

I had to adapt a bit the package.json files to make the packages public for the tests, but I think it is not necessary for the real packages.

pascalgrimaud:
You go too fast for me :D
What do you think about using GitHub actions for release, instead of Azure Pipelines ?

PR #268, opened 2019-09-24
Upgrade chevrotain

clement26695:
I started working on Chevrotain upgrade. It seems that it works with these changes, but the way I implemented it should be improved.

@bd82 if you could review this PR and give me your feedback, it would be great !

I especially modified some "is..." rules to make the tests pass and removed some errors throwing in the process, so I guess something should be improved here.


packages/java-parser/src/productions/expressions.js
```diff
@@ -203,10 +203,14 @@ function defineRules($, t) {
...
 
   $.RULE("primaryPrefix", () => {
-    let isCastExpression = false;
-    if (tokenMatcher($.LA(1).tokenType, t.LBrace)) {
-      isCastExpression = this.BACKTRACK_LOOKAHEAD($.isCastExpression);
-    }
+    const isCastExpression = $.ACTION(() => {
+      let isCastExpression = false;
+      if (tokenMatcher($.LA(1).tokenType, t.LBrace)) {
+        isCastExpression = this.BACKTRACK_LOOKAHEAD($.isCastExpression);
```
bd82: it may suffix to only wrap
this.BACKTRACK_LOOKAHEAD here with $ACTION.
Maybe you could even put the $.ACTION inside the thisBACKTRACK_LOOKAHEAD method?

packages/java-parser/src/productions/expressions.js
```diff
@@ -305,14 +309,18 @@ function defineRules($, t) {
...
-      isRefTypeInMethodRef = this.BACKTRACK_LOOKAHEAD($.isRefTypeInMethodRef);
-    }
+    const isRefTypeInMethodRef = $.ACTION(() => {
+      let isRefTypeInMethodRef = false;
+      // Performance optimization, only perform this backtracking when a '<' is found
+      // TODO: performance optimization evaluation: avoid doing this backtracking for every "<" encountered.
+      //       we could do it once (using global state) per "fqnOrRefType"
+      // We could do it only once for
+      if (tokenMatcher($.LA(1).tokenType, t.Less)) {
+        isRefTypeInMethodRef = this.BACKTRACK_LOOKAHEAD($.isRefTypeInMethodRef);
```
bd82: same as above maybe too much is wrapped in $.ACTION
may have small runtime performance impact.

packages/java-parser/src/productions/expressions.js
```diff
@@ -604,21 +614,19 @@ function defineRules($, t) {
...
-      );
-      const nextTokTypes = firstUnaryExpressionNotPlusMinus.map(
-        x => x.nextTokenType
-      );
-      // uniq
-      firstForUnaryExpressionNotPlusMinus = nextTokTypes.filter(
-        (v, i, a) => a.indexOf(v) === i
-      );
-    }
+    const firstUnaryExpressionNotPlusMinus = this.computeContentAssist(
```
bd82: we dont want to compute this data every time! it is a slow operation

packages/java-parser/src/productions/expressions.js
```diff
@@ -653,10 +661,11 @@ function defineRules($, t) {
     }
 
     // in the middle of a "classReferenceType"
-    $.OPTION2(() => {
+    try {
```
bd82: i don't get this

packages/java-parser/src/productions/expressions.js
clement26695: concerning this one, I tried to keep it outside of the isReferenceTypeCastExpression rule, but it was not giving me the correct infos... So I'm not sure what to do here

packages/java-parser/src/productions/packages-and-modules.js
```diff
@@ -218,31 +222,30 @@ function defineRules($, t) {
       return false;
     });
 
-    try {
-      // the "{importDeclaration}" is a common prefix
-      $.MANY(() => {
+    let continueLoop1 = true;
+    while (continueLoop1) {
```
bd82: why do we need a `while`?

clement26695: I'm not sure :sweat_smile: 

That is why I wanted you to review this : there was a $.MANY before, which was causing an error, so I tried replacing it with this hack, which does not look good at all !

bd82: i will try to checkout your branch

packages/java-parser/src/productions/expressions.js
clement26695: Same as below, the OPTION was throwing an error, so I tried replacing it with this hack

packages/java-parser/src/productions/packages-and-modules.js
bd82: i do not get an error during constructor time.
When did your error occur? perhaps you already fixed it?

bd82:
## Perhaps in a different change:
Try to remove `ignoredIssues` property from the Parser's config.
Most ambiguities that are resolved by `GATE` are auto magically ignored.

You will still get (fewer) errors regarding ambiguities because the new ambiguity ignoring is more granular. Those errors should be inspected, either we already handle those ambiguities in ways other than GATE, in which case they need to be ignored explicitly on the DSL method (new options).
Or perhaps this will cause us to detect some new edge case bugs.

See: https://github.com/SAP/chevrotain/issues/869

bd82:
Remove "hacky" fix for CST Building while backtracking.
See: https://github.com/SAP/chevrotain/issues/789

packages/java-parser/src/productions/expressions.js
clement26695: I just tested with keeping the original code, and it fails when trying to parse

```java
public class DispatcherServletsMappingDescriptionProvider {

	private static <T> List<T> describe() {
		for (HandlerMappingDescriptionProvider<?> descriptionProvider : descriptionProviders) {
			
		}
	}
}
```

with `compilationUnit` as entryPoint in the `single-sample-runner.js` script. I got this error:

```
/home/cdessoude/Documents/workplace/prettier-java/node_modules/chevrotain/lib/src/parse/parser/traits/recognizer_engine.js:188
            throw e;
            ^

TypeError: Cannot read property 'call' of undefined
    at JavaParser.RecognizerEngine.optionInternalLogic (/home/cdessoude/Documents/workplace/prettier-java/node_modules/chevrotain/lib/src/parse/parser/traits/recognizer_engine.js:227:27)
    at JavaParser.RecognizerEngine.optionInternal (/home/cdessoude/Documents/workplace/prettier-java/node_modules/chevrotain/lib/src/parse/parser/traits/recognizer_engine.js:196:25)
    at JavaParser.RecognizerApi.OPTION2 (/home/cdessoude/Documents/workplace/prettier-java/node_modules/chevrotain/lib/src/parse/parser/traits/recognizer_api.js:108:21)
    at JavaParser.$.RULE (/home/cdessoude/Documents/workplace/prettier-java/packages/java-parser/src/productions/expressions.js:654:7)
    at JavaParser.invokeRuleWithTry (/home/cdessoude/Documents/workplace/prettier-java/node_modules/chevrotain/lib/src/parse/parser/traits/recognizer_engine.js:119:33)
    at JavaParser.wrappedGrammarRule [as isRefTypeInMethodRef] (/home/cdessoude/Documents/workplace/prettier-java/node_modules/chevrotain/lib/src/parse/parser/traits/recognizer_engine.js:133:38)
    at JavaParser.BACKTRACK_LOOKAHEAD.ACTION (/home/cdessoude/Documents/workplace/prettier-java/packages/java-parser/src/parser.js:182:27)
    at JavaParser.RecognizerApi.ACTION (/home/cdessoude/Documents/workplace/prettier-java/node_modules/chevrotain/lib/src/parse/parser/traits/recognizer_api.js:21:21)
    at JavaParser.BACKTRACK_LOOKAHEAD (/home/cdessoude/Documents/workplace/prettier-java/packages/java-parser/src/parser.js:175:17)
    at JavaParser.$.RULE (/home/cdessoude/Documents/workplace/prettier-java/packages/java-parser/src/productions/expressions.js:314:35)
```

clement26695:
> Try to add $.ACTION directly inside this.BACKTRACK_LOOKAHEAD?

Done ! 

I will try to delete the `ignoredIssues` afterwards, when the changes in the parser look good enough

packages/java-parser/src/productions/packages-and-modules.js
clement26695: It seems working for this one, I reverted the changes. Not sure why I modified it in the first place

packages/java-parser/src/productions/classes.js
```diff
@@ -662,24 +662,19 @@ function defineRules($, t) {
...
         }
-      });
-    } catch (e) {
-      if (isRecognitionException(e)) {
-        // TODO: add original syntax error?
-        throw "Cannot Identify if the <TypeDeclaration> is a <ClassDeclaration> or an <InterfaceDeclaration>";
       } else {
-        throw e;
+        continueLoop = false;
       }
```
clement26695: It is the same as the OPTION in `isReferenceTypeCastExpression`. If I try to keep the original code, it fails even on:

```java
class T {}
```

with `compilationUnit` as entryPoint in single-sample-runner.js

packages/java-parser/src/productions/expressions.js
clement26695: For instance, when trying to parse
 ```java
PlayerDetails other = (PlayerDetails) obj;
```
with `localVariableDeclarationStatement` as entryPoint in `single-sample-runner.js`. I got this:

```
/home/cdessoude/Documents/workplace/prettier-java/packages/java-parser/src/index.js:44
    throw Error(
    ^

Error: Sad sad panda, parsing errors detected in line: 2, column: 39!
Expecting --> ';' <-- but found --> 'obj' <--!
	->localVariableDeclarationStatement
    at Object.parse (/home/cdessoude/Documents/workplace/prettier-java/packages/java-parser/src/index.js:44:11)
    at Object.<anonymous> (/home/cdessoude/Documents/workplace/prettier-java/packages/java-parser/scripts/single-sample-runner.js:12:16)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:831:12)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:622:3)
```

When I print the `firstForUnaryExpressionNotPlusMinus` before consuming the `t.LBrace`, I only have a `UnaryPrefixOperatorNotPlusMinus` and a `UnarySuffixOperator` when I had much more tokens with my change

File: packages/java-parser/src/productions/expressions.js (hunk 4 above)
Comment (bd82) [reply packages/java-parser/src/productions/expressions.js]: maybe you need to only compute this after the recording phase has finished?

Comment (bd82) [reply packages/java-parser/src/productions/expressions.js]: you could move this calculation to the last step of the constructor for example

File: packages/java-parser/src/productions/classes.js (hunk 6 above)
Comment (bd82) [reply packages/java-parser/src/productions/classes.js]: I will try to reproduce the error when running the tests

Comment (bd82) [reply packages/java-parser/src/productions/classes.js]: The recording logic gets broken because of an early "return false".
Make sure the `isXXXXX` will not return early during recording phase.
You can use [this flag](https://sap.github.io/chevrotain/documentation/6_5_0/classes/baseparser.html#recording_phase) to assist you.

Basically you want all the DSL methods in the Rules to be invoked during "recording"
If you debug such an "isABC" rule you can see it exits early without getting to the MANY/OPTION and therefor the lookahead function for that rule would not be created
leading to `TypeError: Cannot read property 'call' of undefined`

Comment (clement26695) [reply packages/java-parser/src/productions/classes.js]: Great ! I refactored the function to avoid early returns

File: packages/java-parser/src/productions/interfaces.js [hunk 7]
```diff
@@ -239,17 +241,22 @@ function defineRules($, t) {
...
+          },
+          {
+            ALT: () => {
+              /* empty normal annotation contents */
+            }
           }
-        }
-      ]);
+        ],
+        IGNORE_AMBIGUITIES: true
```
clement26695: Ambiguous Alternatives Detected: <1 ,2> in <OR> inside <annotation> Rule,
<Identifier, '='> may appears as a prefix path in all these alternatives.

It is because an elementValue could be a binaryExpression and could start with Identifier and '='. Do you think we must use a GATE or backtracking here ?

packages/java-parser/src/productions/blocks-and-statements.js
```diff
@@ -28,12 +28,16 @@ function defineRules($, t) {
...
+
+    const isClassDeclaration = this.BACKTRACK_LOOKAHEAD($.isClassDeclaration);
+
     $.OR([
       {
         GATE: () => isLocalVariableDeclaration,
         ALT: () => $.SUBRULE($.localVariableDeclarationStatement)
       },
       {
+        GATE: () => isClassDeclaration,
```
clement26695: Not sure if this is needed ? We got otherwise:
```
 Ambiguous Alternatives Detected: <2 ,3> in <OR> inside <blockStatement> Rule,
<'@', Identifier> may appears as a prefix path in all these alternatives.
```

because a statement can have annotations

packages/java-parser/src/productions/blocks-and-statements.js
```diff
@@ -57,10 +61,13 @@ function defineRules($, t) {
...
-    $.OR([
-      { ALT: () => $.SUBRULE($.unannType) },
-      { ALT: () => $.CONSUME(t.Var) }
-    ]);
+    $.OR({
+      DEF: [
+        { ALT: () => $.SUBRULE($.unannType) },
+        { ALT: () => $.CONSUME(t.Var) }
+      ],
+      IGNORE_AMBIGUITIES: true
```
clement26695: Not sure how to enforce not using 'var' in unannType

clement26695:
> Remove "hacky" fix for CST Building while backtracking.
> See: [SAP/chevrotain#789](https://github.com/SAP/chevrotain/issues/789)

Concerning this one, It would be great if @Shaolans could have a look, as he added the "prettier-ignore" logic here

packages/java-parser/src/productions/expressions.js
clement26695: You are talking about the `firstUnaryExpressionNotPlusMinus` issue, right (not about the `$.OPTION`) ? You would try to compute it after the this.performSelfAnalysis() in the JavaParser constructor ?

Shaolans:
> Concerning this one, It would be great if @Shaolans could have a look, as he added the "prettier-ignore" logic here

I will take a look

packages/java-parser/src/productions/blocks-and-statements.js
bd82: makes sense

packages/java-parser/src/productions/blocks-and-statements.js
bd82: can var be used as an identifier? is that the cause of the ambiguity?

packages/java-parser/src/productions/expressions.js
```diff
@@ -166,6 +169,8 @@ function defineRules($, t) {
           }
         },
         {
+          // Remove ambiguity with first OR option
+          GATE: () => !tokenMatcher($.LA(1).tokenType, t.Instanceof),
```
bd82: is the ambiguity is resolved by the order (instanceOf is first) you don't need a GATE
rather you can ignore the ambiguities in this specific ALT

packages/java-parser/src/productions/interfaces.js
bd82: if you use 3 or 4 tokens lookahead would that resolve in this case?
because you can increase the maxLookahead for this specific rule...
I don't remember the specific grammar rules, does the `list` alternative always have a comma somewhere? how far ahead is that comma?

packages/java-parser/src/productions/types-values-and-variables.js
```diff
@@ -56,24 +56,27 @@ function defineRules($, t) {
...
-        ALT: () => {
-          $.SUBRULE($.classOrInterfaceType);
-          $.OPTION(() => {
-            $.SUBRULE2($.dims);
-          });
-        }
-      }
-    ]);
+      ],
+      IGNORE_AMBIGUITIES: true // annotation prefix was extracted to remove ambiguities
```
bd82: good comment

packages/java-parser/src/productions/blocks-and-statements.js
bd82: If that is the case (I have some fuzzy memory of limitations for the `var` keyword.
You may be able to define a IdentifierButNotVar Token using token categories.

clement26695: Yes, that's right. I think I will try to improve this in a following PR, adding a token and change some parser logic seems out of the scope of this PR

bd82: Yes definitively out of scope

packages/java-parser/src/productions/expressions.js
bd82: Yes you are correct I must have replied in the wrong thread 😄

packages/java-parser/src/productions/expressions.js
clement26695: I'm not sure how to proceed on this one: I tried to extract the logic into one method and call it at the end of the parser constructor, but didn't manage to make it work. If I extract the logic, I got this error `Error: Rule ->unaryExpressionNotPlusMinus<- does not exist in this grammar.`, so I guess I am not doing the right thing 😄

bd82:
```javascript
    // ...
    // prevents runtime error during the recording phase
    this.firstForUnaryExpressionNotPlusMinus= [];
    this.performSelfAnalysis();

    const firstUnaryExpressionNotPlusMinus = this.computeContentAssist(
      "unaryExpressionNotPlusMinus",
      []
    );
    const nextTokTypes = firstUnaryExpressionNotPlusMinus.map(
      x => x.nextTokenType
    );
    // uniq
    this.firstForUnaryExpressionNotPlusMinus = nextTokTypes.filter(
      (v, i, a) => a.indexOf(v) === i
    );
```

```javascript
  $.RULE("isReferenceTypeCastExpression", () => {
    $.CONSUME(t.LBrace);
    $.SUBRULE($.referenceType);
    $.MANY(() => {
      $.SUBRULE($.additionalBound);
    });
    $.CONSUME(t.RBrace);
    const firstTokTypeAfterRBrace = this.LA(1).tokenType;

    return (
      this.firstForUnaryExpressionNotPlusMinus.find(tokType =>
        tokenMatcher(firstTokTypeAfterRBrace, tokType)
      ) !== undefined
    );
  });
```

packages/java-parser/src/productions/interfaces.js
clement26695: Unfortunately, it is not enough to increase the maxLookahead for this rule. The list alternative always has a comma somewhere, but as the elementValue could be an expression, it could have several annotations and thus we cannot look ahead at a specific step.

packages/java-parser/src/parser.js
```diff
@@ -68,7 +68,9 @@ class JavaParser extends Parser {
     blocksStatements.defineRules.call(this, $, t);
     expressions.defineRules.call(this, $, t);
 
+    this.firstForUnaryExpressionNotPlusMinus = [];
     this.performSelfAnalysis();
+    expressions.computeFirstForUnaryExpressionNotPlusMinus.call(this);
```
bd82: `this.firstForUnaryExpressionNotPlusMinus = expressions.computeFirstForUnaryExpressionNotPlusMinus.call(this);`

Would be clearer in terms of tacking state changes

clement26695:
@bd82: Do you think it can be merged ? We could do the rest of the work (removing "hacky" fix for CST Building while backtracking & working on IGNORE_AMBIGUITIES cases in follow-up PRs)

bd82:
> @bd82: Do you think it can be merged ? We could do the rest of the work (removing "hacky" fix for CST Building while backtracking & working on IGNORE_AMBIGUITIES cases in follow-up PRs)

Yes, this looks good 👍

PR #273, opened 2019-10-01
Feature: Organize imports according to Google style

### LINKED ISSUE #201: Organize imports ?
Should we ordered imports as in the following guides ?

https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing

or 

https://github.com/spring-projects/spring-framework/wiki/Code-Style

clement26695:
Fix #201


jhaber:
Is it worth adding a test for a class with only static imports, and a class with only non-static imports?

jhaber:
👍 I was mainly thinking to verify that there isn't extra whitespace in those cases (both in the current impl but also for anyone who refactors things in the future)

packages/prettier-plugin-java/src/printers/printer-utils.js
```diff
@@ -531,6 +531,63 @@ function isStatementEmptyStatement(statement) {
...
+      compareFqn(
+        first.children.packageOrTypeName[0],
+        second.children.packageOrTypeName[0]
+      )
+    );
+    nonStaticImports.sort((first, second) =>
+      compareFqn(
+        first.children.packageOrTypeName[0],
+        second.children.packageOrTypeName[0]
+      )
```
murdos: Maybe you could factor this by declaring the comparator in a variable and then use it?
```
const comparator = (first, second) =>
      compareFqn(
        first.children.packageOrTypeName[0],
        second.children.packageOrTypeName[0]
      );
staticImports.sort(comparator);
nonStaticImports.sort(comparator);
```

PR #280, opened 2019-10-08
Fix formatComment to keep stable reformating and add comments unit tests

### LINKED ISSUE #279: Multiline comment formatting not idempotent
I'm seeing an issue with multiline comments getting extra indentation each time they get run through prettier-java

Input:
```java
public class Example {
  
  public void method() {
    /*
      multiline comment
     */
  }
}
```
Output: (after running prettier ~5 times)
```java
public class Example {

  public void method() {
    /*
                multiline comment
   */
  }
}
```

Each run of prettier seems to indent the comment a little more.

The issue doesn't manifest if each line of the comment is prefixed with an asterisk (although the formatting in this case still looks a little bit off), for example:

Input:
```java
public class Example {

  public void method() {
    /*
     * multiline comment
     */
  }
}
```

Output:
```java
public class Example {

  public void method() {
    /*
   * multiline comment
   */
  }
}
```

clement26695:
Fix #279


jhaber:
👍 thanks for the quick fix

lppedd:
@clement26695 another suggestion. Take for example function definition.
```
public Long create(final InlineObject7 /* BackupItem */ dto) {
```
Currently it gets formattet as
```
public Long create(final InlineObject7 /* BackupItem */dto) {
```
So the space before `dto` is removed.  
It would be better, if possible, to keep it.

clement26695:
I will keep it in mind @lppedd ! I prefer to merge this as it is better than that we have right now but will try to find time to improve this

PR #289, opened 2019-11-13
Formatter off on

### LINKED ISSUE #287: Handle @formatter:off and @formatter:on ?
`// @formatter:off` and `// @formatter:on` are commonly used and recognized by Eclipse and IntelliJ.
I wonder if prettier-java should handle them too, a bit like `// prettier-ignore`

nonomoho:
fix #287


pascalgrimaud:
good job @nonomoho !

packages/java-parser/src/comments.js
```diff
@@ -155,6 +155,61 @@ function shouldIgnore(node, comments, ignoredNodes) {
   }
 }
 
+function filterFormatterOffOn(comments) {
+  return [...comments].filter(comment =>
+    comment.image.match(
+      /(\/\/(\s*)@formatter:off(\s*))|(\/\*(\s*)@formatter:off(\s*)\*\/)|(\/\/(\s*)@formatter:on(\s*))|(\/\*(\s*)@formatter:on(\s*)\*\/)/gm
```
Shaolans: I think you can reduce the regex with this ```(\/\/(\s*)@formatter:(off|on)(\s*))|(\/\*(\s*)@formatter:(off|on)(\s*)\*\/)```

packages/java-parser/src/comments.js
```diff
@@ -155,6 +155,61 @@ function shouldIgnore(node, comments, ignoredNodes) {
...
+      }
+    } else {
+      if (!isCurrentCommentOff) {
+        paired.on = comment;
+        pairs.push(paired);
+        paired = {};
+      }
+    }
+
+    if (index === array.length - 1 && isCurrentCommentOff) {
```
clement26695: I think this can move outside the loop and the check replaced by 
```js
if (comments.length > 0 && isCurrentCommentOff) {
```

packages/java-parser/src/comments.js
```diff
@@ -155,6 +155,61 @@ function shouldIgnore(node, comments, ignoredNodes) {
...
+    )
+  );
+}
+
+function matchFormatterOffOnPair(comments) {
+  let isPreviousCommentOff = false;
+  let isCurrentCommentOff = true;
+  const pairs = [];
+  let paired = {};
+  comments.forEach((comment, index, array) => {
```
clement26695: `index` and `array` can be removed

packages/java-parser/src/comments.js
```diff
@@ -155,6 +155,61 @@ function shouldIgnore(node, comments, ignoredNodes) {
...
+
+    isPreviousCommentOff = isCurrentCommentOff;
+  });
+
+  return pairs;
+}
+
+function shouldNotFormat(node, commentPairs) {
+  const matchingPair = _.find(
+    commentPairs.reverse(),
```
clement26695: Be aware that this method is called for each node of the file, so the reverse method is called a lot. You can maybe use `findLast` from lodash to avoid calling it, or reverse it directly in the `setOffOnComments` to do it once

packages/java-parser/src/comments.js
```diff
@@ -155,6 +155,61 @@ function shouldIgnore(node, comments, ignoredNodes) {
...
+}
+
+function matchFormatterOffOnPair(comments) {
+  let isPreviousCommentOff = false;
+  let isCurrentCommentOff = true;
+  const pairs = [];
+  let paired = {};
+  comments.forEach((comment, index, array) => {
+    isCurrentCommentOff =
+      comment.image.slice(comment.image.length - 3) === "off";
```
clement26695: You can use `.slice(-3)` 😄

nonomoho: I do that tomorrow!

packages/java-parser/src/comments.js
nonomoho: I agree, you're right!

clement26695:
Would it be possible to add some tests at some other places than just before classes definition ? Before methods, binary operations and so on

clementdessoude:
Closed with #323

PR #295, opened 2019-11-22
performance(parser) - Optimize Parser Init

### LINKED ISSUE #269: Parser Initilization Time
It seems to take well over 100ms on my machine for the Parser to initialize.
There is now an official guide how to speed the initialization time:
https://sap.github.io/chevrotain/docs/guide/initialization_performance.html#enabling-initialization-performance-tracing

Please report on the results, additional upgrades can be implemented for faster init time as needed...

Also note that enabling the init perf tracing affects the initialization time due to many (synchronized) console printing. So the real performance number is best achieved with a simple
timer. 

```
const start = new Date.getTime();
// do smt
const end = new Date.getTime();
const total = end - start;
```

bd82:
Optimize Parser (And Lexer) initialization performance by skipping validation
during productive flows.
The validations are still executed during testing flows using
a mocha configuration file that adds an env variable.

On my Machine the Parser's init time went from 160ms->90ms
And the Lexer's init time went from 30ms->20ms.

Fixes #269

### DIFF
--- a/packages/java-parser/src/parser.js
+++ b/packages/java-parser/src/parser.js
@@ -11,6 +11,7 @@ const arrays = require("./productions/arrays");
 const blocksStatements = require("./productions/blocks-and-statements");
 const expressions = require("./productions/expressions");
 const { shouldIgnore } = require("./comments");
+const { getSkipValidations } = require("./utils");
 
 /**
  * This parser attempts to strongly align with the specs style at:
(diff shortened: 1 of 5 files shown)


packages/java-parser/src/parser.js
```diff
@@ -38,10 +39,12 @@ const { shouldIgnore } = require("./comments");
...
   constructor() {
     super(allTokens, {
-      // TODO: performance: maxLookahead = 1 may be faster, but could we refactor the grammar to it?
-      //       and more importantly, do we want to?
+      // TODO: Try to Specify max lookahead 2 only where needed
+      //       and use `1` by default
       maxLookahead: 2,
-      nodeLocationTracking: "full"
+      nodeLocationTracking: "full",
+      // traceInitPerf: 2,
```
clement26695: Is this comment wanted ?

clement26695:
LGTM.

Just for info: how do you measure the performance ?

bd82:
> Just for info: how do you measure the performance ?

```javascript
const start = new Date().getTime();
const parser = new Parser();
const end = new Date().getTime();
console.log(end - start);
```

Issue #291, opened 2019-11-15
Enforce modifier order


jhaber:
I think it would be in-scope for prettier-java to rewrite class, field, and method modifiers to match the order specified in the Java Language Specification. Here's the order that checkstyle uses:
https://checkstyle.sourceforge.io/config_modifier.html#ModifierOrder

I was testing with this code:
```java
public class Test {

  static public void main(String... args) {}
}
```
So the desired output would be:
```java
public class Test {

  public static void main(String... args) {}
}
```

jhaber:
If you agree it's appropriate for prettier-java to handle this, I can attempt to send a PR in the next few weeks

clement26695:
Good idea ! Please go ahead: if you need any help, let me know !

clementdessoude:
Closed with #302

PR #302, opened 2019-11-28
Sort modifiers

jhaber:
Related to #291 

I had some trouble getting access to the actual modifier name ("public", "static", "private", etc.) and I don't write much javascript so any feedback on the code is welcome. 

There is an existing `sortModifiers` function that everything goes through, however it seems to just separate leading annotations from all other annotations/modifiers. I tried to preserve the behavior of this function, however it conflicts a bit with the new modifier sorting behavior. 

For example, if you have this code:
`final @Annotation private String field;`

Currently, prettier-java leaves this code alone. However, we want to swap the order of the `final` and `private` modifiers, but it's not clear what should happen to `@Annotation` when we do this. As currently written, this PR moves all such annotations to the end, and would output:
`private final @Annotation String field;`

Some other options that come to mind:
- unconditionally move annotations ahead of modifiers, which means you can't write a method signature like `public @Nullable String myMethod()`
- move annotations to the front unless they come after all modifiers, which still allows method signatures like `public @Nullable String myMethod()`

### DIFF
--- a/packages/prettier-plugin-java/docs/modifiers.md
+++ b/packages/prettier-plugin-java/docs/modifiers.md
@@ -39,7 +39,7 @@ private static final String S = "abc";
 Input:
 
 ```java
-protected abstract class C {}
+abstract protected class C {}
 ```
 
 Output:
--- a/packages/prettier-plugin-java/docs/modifiers.md
+++ b/packages/prettier-plugin-java/docs/modifiers.md
@@ -1,0 +1,100 @@
+# Modifier formatting
+
+### Modifier sorting
+
+Class, method, and field modifiers will be sorted according to the order in the JLS (sections [8.1.1](https://docs.oracle.com/javase/specs/jls/se11/html/jls-8.html#jls-8.1.1), [8.3.1](https://docs.oracle.com/javase/specs/jls/se11/html/jls-8.html#jls-8.3.1), [8.4.3](https://docs.oracle.com/javase/specs/jls/se11/html/jls-8.html#jls-8.4.3), and [9.4](https://docs.oracle.com/javase/specs/jls/se11/html/jls-9.html#jls-9.4)).
+The overall ordering is:
+
+- public
+- protected
+- private
+- abstract
+- default
+- static
+- final
+- transient
+- volatile
+- synchronized
+- native
+- strictfp
+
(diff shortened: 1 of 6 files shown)


clement26695:
Just maybe one thing to add:would it be possible to add a docs section in the prettier-plugin-java package with a description of this sorting behavior as it might not seem directly intuitive?

jhaber:
Sure thing, is there an existing place where that sort of documentation should go or should I add a docs folder within `packages/prettier-plugin-java`?

clement26695:
No, there is no existing place yet. Can you create the folder please?

jhaber:
I added `packages/prettier-plugin-java/docs/modifiers.md`, let me know how it looks to you

clement26695:
LGTM! I will have a final review before merging it but it should be good

murdos:
> 
> 
> LGTM. I don't think there is a better way for accessing the type of modifier. I'm not sure either about what should be done with annotations but this one looks good to me. Do you have an opinion about this @murdos?

For a field I would expect `final @Annotation private String field;` to become
```java
@Annotation
private final String field;
```

but as @jhaber said, for a method, we should correctly handle annotation attached to return type of a method : `public @Nullable String myMethod()`

jhaber:
Yeah, thinking about it more the concern with preserving annotation placement only applies to methods where you have ambiguity about method annotations vs. return type annotations. So I would propose updating the logic so that annotations unconditionally get moved to the front and on their own lines for  classes and fields. For methods, the same will happen unless the annotation comes after all other modifiers

jhaber:
I've updated the code and documentation to implement the approach described in my previous comment. However, this is causing a test failure on `comments/class/_input.java`:
```java
private transient @MonotonicNonNull ColumnMap columnMap;
```

Previously this was being left alone, but is now getting re-written to:
```java
@MonotonicNonNull
private transient ColumnMap columnMap;
```

This matches my expectation, so  I updated `comments/class/_output.java` to make the test pass. Let me know if that's the wrong approach to fixing the issue.

clementdessoude:
Thank you again @jhaber !

jhaber:
Glad I could help a tiny bit with this awesome project 😄

PR #310, opened 2019-12-05
Fix/parser variable arity

clementdessoude:
Fix last parser identified bugs with variable Arity parameters (see #139)

### DIFF
--- a/packages/java-parser/src/productions/types-values-and-variables.js
+++ b/packages/java-parser/src/productions/types-values-and-variables.js
@@ -136,12 +136,15 @@ function defineRules($, t) {
     });
     $.CONSUME(t.LSquare);
     $.CONSUME(t.RSquare);
-    $.MANY2(() => {
-      $.MANY3(() => {
-        $.SUBRULE2($.annotation);
-      });
-      $.CONSUME2(t.LSquare);
-      $.CONSUME2(t.RSquare);
+    $.MANY2({
+      GATE: () => $.BACKTRACK_LOOKAHEAD($.isDims),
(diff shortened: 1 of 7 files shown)


packages/java-parser/src/productions/types-values-and-variables.js
```diff
@@ -136,12 +136,15 @@ function defineRules($, t) {
...
     $.CONSUME(t.LSquare);
     $.CONSUME(t.RSquare);
-    $.MANY2(() => {
-      $.MANY3(() => {
-        $.SUBRULE2($.annotation);
-      });
-      $.CONSUME2(t.LSquare);
-      $.CONSUME2(t.RSquare);
+    $.MANY2({
+      GATE: () => $.BACKTRACK_LOOKAHEAD($.isDims),
```
bd82: this GATE seems too restrictive, what if it starts with annotation?

clementdessoude: I don't really understand: if it start with annotations, the Gate will be OK, as the `isDims` rule start with MANY(annotation). Or am I missing something? The name isDims is maybe a bit misleading

clementdessoude: I don't really understand: if it start with annotations, the Gate will be OK, as the `isDims` rule start with MANY(annotation). Or am I missing something? The name isDims is maybe a bit misleading

bd82: I missed that isDims also checks annotations

PR #318, opened 2019-12-06
fix: ternary expression should break before `?` and `:`

### LINKED ISSUE #312: Mutli-line ternary expression indentation seems inconsistent with if/else formatting
Multi-line/wrapped ternary expressions are currently formatted like:
```java
String wasExpectedToDropString = expectedToDrop ? "was expected to drop"
  : "was NOT expected to drop";
```

If else expressions are formatted like:

```java
if (expectedToDrop) {
  wasExpectedToDropString = "was expected to drop";
} else {	
  myString = "was NOT expected to drop";
}
```

In my opinion it would make sense to have multi-line ternary expressions similarly formatted like:

```java
String wasExpectedToDropString = expectedToDrop 
  ? "was expected to drop"
  : "was NOT expected to drop";
```

This feels consistent to me with the if/else formatting, other boolean formatting, etc. 

Apologies if this was already addressed somewhere, but I couldn't find any issue explicitly making a decision on this, or raising this question previously.

clementdessoude:
Fix #312

### DIFF
--- a/packages/prettier-plugin-java/src/printers/expressions.js
+++ b/packages/prettier-plugin-java/src/printers/expressions.js
@@ -131,20 +131,16 @@ class ExpressionsPrettierVisitor {
       const expression1 = this.visit(ctx.expression[0]);
       const expression2 = this.visit(ctx.expression[1]);
 
-      return group(
-        rejectAndConcat([
-          group(
-            rejectAndConcat([
+      return indent(
+        group(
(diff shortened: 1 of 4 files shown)


jhaber:
👍 thanks for the quick fix!

natdempk:
Thanks for the fix here as well! 🙂

PR #319, opened 2019-12-07
Feature/enums improvements

### LINKED ISSUE #313: Enum trailing commas get included on the same line as the semicolon ending the line
Enums that get split across multiple lines can lead to the last enum in the series having a `, ;` on the last line. If a `,` is specified after the last enum value that should be used as an indicator to create another newline and to place the `;` on that newline, or the `, ` should be dropped entirely and it should just become a `;` after the last enum value. 

I could see arguments for either option, but I'd prefer placing the `;` on a newline as this preserves version control history for the line with the previous enum value when a new value is eventually added.

Input:
```java
enum MyEnum {
  AAAAAAAA("aaaaaa"),
  BBBBBBBBBBB("bbbbbbbbbb"),
  CCCCCCCCCC("ccccccccc"),
  DDDDDDDDD("ddddddddddddddddddddddddd"),
  EEEEEEE("eeeeeeeeee"), 
  ;
}
```

Output:
```java
enum MyEnum {
  AAAAAAAA("aaaaaa"),
  BBBBBBBBBBB("bbbbbbbbbb"),
  CCCCCCCCCC("ccccccccc"),
  DDDDDDDDD("ddddddddddddddddddddddddd"),
  EEEEEEE("eeeeeeeeee"), ;
}
```

Expected Output:
```java
enum MyEnum {
  AAAAAAAA("aaaaaa"),
  BBBBBBBBBBB("bbbbbbbbbb"),
  CCCCCCCCCC("ccccccccc"),
  DDDDDDDDD("ddddddddddddddddddddddddd"),
  EEEEEEE("eeeeeeeeee"),
  ;
}
```

### LINKED ISSUE #314: Enums end up on the same line squashing git diffs and changing formatting as values are added
Similar to https://github.com/jhipster/prettier-java/issues/264 and https://github.com/jhipster/prettier-java/issues/313

Enums will get collapsed onto a single line, and then later as more are added, spread back to a single line per enum. Eg. if I have 

```java
enum ChangeType { 
  MIGRATION,
  CREATED,
}
```

Initially this will get turned into:

```java
enum ChangeType { 
  MIGRATION, CREATED,
}
```

But later if I add more values over time, at some point it will get turned back into something like 
```java
enum ChangeType { 
  MIGRATION,
  CREATED,
  MORE,
  ENUM,
  VALUES,
  OVER,
  TIME,
  AS,
  THIS,
  GROWS,
}
```

This means that all the VCS history for why those enum values were added linewise is lost over time, but if we had just maintained the one-per-line formatting over time this would not be an issue, and the formatting would remain consistent as values are added. 

Are there strong arguments against having enums always be one-per-line? IMO it helps with VCS history and provides a more consistent format everywhere enums are used.

clementdessoude:
Fix #314 and fix #313 

I didn't implement a `--trailing-comma` comma for the moment. It could be added in the feature, but I don't think it is a top priority for the moment

### DIFF
--- a/packages/prettier-plugin-java/src/printers/classes.js
+++ b/packages/prettier-plugin-java/src/printers/classes.js
@@ -674,7 +674,9 @@ class ClassesPrettierVisitor {
 
   enumConstantList(ctx) {
     const enumConstants = this.mapVisit(ctx.enumConstant);
-    const commas = ctx.Comma ? ctx.Comma.map(elt => concat([elt, line])) : [];
+    const commas = ctx.Comma
+      ? ctx.Comma.map(elt => concat([elt, hardline]))
+      : [];
 
     return group(rejectAndJoinSeps(commas, enumConstants));
   }
--- a/packages/prettier-plugin-java/src/printers/classes.js
+++ b/packages/prettier-plugin-java/src/printers/classes.js
@@ -16,11 +16,7 @@ const {
   isStatementEmptyStatement
 } = require("./printer-utils");
 const { concat, join, group, indent } = require("./prettier-builder");
-const {
-  printTokenWithComments,
-  getTokenLeadingComments,
-  getTokenTrailingComments
-} = require("./comments");
+const { printTokenWithComments } = require("./comments");
 
 class ClassesPrettierVisitor {
   classDeclaration(ctx) {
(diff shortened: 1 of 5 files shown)


jhaber:
:+1: this looks great

natdempk:
Thanks for the fix here!

PR #330, opened 2019-12-12
Sort imports naively

### LINKED ISSUE #324: Order of import statements
In `0.5.1`, the import statements are ordered ignoring the letter cases, e.g.:

```java
import java.util.ArrayList;
import java.util.function.Consumer;
import java.util.List;
import java.util.Map;
```

It would be good to have the uppercase letter rank first like this

```java
import java.util.ArrayList;
import java.util.List;
import java.util.Map;
import java.util.function.Consumer;
```

jhaber:
Fixes #324 

I believe this matches the Google import sort order
https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing
> Within each block the imported names appear in ASCII sort order. (Note: this is not the same as the import statements being in ASCII sort order, since '.' sorts before ';'.)

This also appears to better conform to the way IntelliJ/Eclipse sort imports. I looked through the tests for Google's import sorter and tried to cover the same edge cases in the tests I added:
https://github.com/google/google-java-format/blob/master/core/src/test/java/com/google/googlejavaformat/java/ImportOrdererTest.java

### DIFF
--- a/packages/prettier-plugin-java/src/printers/printer-utils.js
+++ b/packages/prettier-plugin-java/src/printers/printer-utils.js
@@ -579,11 +579,10 @@ function compareFqn(packageOrTypeNameFirst, packageOrTypeNameSecond) {
 
   const minParts = Math.min(identifiersFirst.length, identifiersSecond.length);
   for (let i = 0; i < minParts; i++) {
-    const identifierComparison = identifiersFirst[i].image.localeCompare(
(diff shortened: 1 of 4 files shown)


clementdessoude:
LGTM. I just have one question: what happen when you have the same class name in two imports with a star in Java? Does it break compilation?

PR #334, opened 2019-12-15
Improve unit tests

clementdessoude:
Create a method to have unit tests on Java code snippets


packages/prettier-plugin-java/test/test-utils.js
```diff
@@ -69,22 +67,22 @@ function testRepositorySample(testFolder, command, args) {
     );
 
     javaSampleFiles.forEach(fileDesc => {
-      it(`prettify ${relative(samplesDir, fileDesc.path)}`, function() {
+      it(`Performs a stable formatting for <${relative(
+        samplesDir,
+        fileDesc.path
+      )}>`, function() {
         this.timeout(5000);
```
bd82: Why 5000ms timeout?

packages/prettier-plugin-java/test/unit-test/snippets/names/methodName/test.spec.js
```diff
@@ -0,0 +1,15 @@
+"use strict";
+
+const { expect } = require("chai");
+const { formatJavaSnippet } = require("../../../../test-utils");
+
+describe("MethodName", () => {
+  it("can format a MethodName", () => {
+    const snippet = "test";
```
bd82: these 5 lines keep repeating, perhaps extract to a utility?

packages/prettier-plugin-java/test/test-utils.js
clementdessoude: I just kept that from what it was before. @Shaolans may have an answer to why it was added in the first place?

bd82: this may be needed for bigger tests, does not seem appropriate for small unit tests

PR #342, opened 2019-12-23
Update Changelog to v0.6.0

clementdessoude:


bd82:
@clementdessoude are you interested in automated change log generation and standardized commit messages?

E.g: https://github.com/SAP/xml-tools/blob/master/packages/ast/CHANGELOG.md

clementdessoude:
I think it can be useful, but I have to take a deeper look

bd82:
o.k. Also have a look here when you have time:
- https://medium.com/angular-in-depth/release-management-in-angular-with-lerna-21b4ab417c59

The repo I linked to in the example above also uses Lerna in **independent** mode
- https://github.com/lerna/lerna#independent-mode

PR #353, opened 2019-12-30
fix: move comments in binaryExpression

### LINKED ISSUE #352: Collapsing comments within expressions
**Prettier-Java 0.5.1**

The following input causes prettier to produce output that is both in correct and that changes when prettier is run again. It seems as though I need to run `prettier --write` 4 times before it reaches the state shown in the output. 

**Input:**

```java
public class InlineComment {

  public boolean method() {
    return "some string value".equals("some string value")
        // This is a very long multiline comment that appears in the middle of an expression:
        // this seems to cause problems
        // for prettier.
        || "some other string value".equals("some other string value");
  }
}
```

**Output:**

```java
public class InlineComment {

  public boolean method() {
    return (
      "some string value".equals("some string value") || // for prettier. // this seems to cause problems // This is a very long multiline comment that appears in the middle of an expression:
      "some other string value".equals("some other string value")
    );
  }
}
```

**Expected behavior:**

The javascript equivalent maintains the long comments:
```js
function method() {
  return (
    "some string value" === "some string value" ||
    // This is a very long multiline comment that appears in the middle of an expression:
    // this seems to cause problems
    // for prettier.
    "some other string value" === "some other string value"
  );
}
```

The seems very reasonable to me. I'd say that whatever happens should require a single execution of prettier.

If comments are still collapse, they should be in the correct order. In such a case, I think it would also be nice to trim the leading `// ` when doing so, but I understand that that may be difficult.

clementdessoude:
Move comments before binaryOperator to the following unaryExpression

Fix #352

By comparison, what is done in Prettier for TS:

```sh
--parser typescript
```

**Input:**
```tsx
a = one || two >> 1 // one
        // two
        // three
        || //four
        // five
        three;
```

**Output:**
```tsx
a =
  one ||
  two >> 1 || // one //four
  // two
  // three
  // five
  three;

```

### DIFF
--- a/packages/prettier-plugin-java/src/printers/expressions.js
+++ b/packages/prettier-plugin-java/src/printers/expressions.js
@@ -15,7 +15,8 @@ const {
   putIntoBraces,
   separateTokensIntoGroups,
   isShiftOperator,
-  isUniqueMethodInvocation
+  isUniqueMethodInvocation,
+  handleComments
 } = require("./printer-utils");
 
 class ExpressionsPrettierVisitor {
@@ -158,6 +159,8 @@ class ExpressionsPrettierVisitor {
   }
 
   binaryExpression(ctx, params) {
+    handleComments(ctx);
+
     const referenceType = this.mapVisit(ctx.referenceType);
     const expression = this.mapVisit(ctx.expression);
     const unaryExpression = this.mapVisit(ctx.unaryExpression);
--- a/packages/prettier-plugin-java/src/printers/printer-utils.js
+++ b/packages/prettier-plugin-java/src/printers/printer-utils.js
@@ -634,6 +634,27 @@ function isUniqueMethodInvocation(primarySuffixes) {
   return count;
 }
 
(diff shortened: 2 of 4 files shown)


clementdessoude:
For the moment, there is still some unexpected behavior. But at least it is stable. For instance:

** Input **
```java
public class InlineComment {

  public boolean method() {
    boolean a = one || two >> 1 // one
        // two
        // three
        || // five
        // four
        three;

        boolean a = one || two >> 1 // one
        // two
        // three
        ||
        three;

        boolean a = one || two >> 1 // one
        // two
        // three
        || three;
  }
}
```

** Output **
```java
public class InlineComment {

  public boolean method() {
    boolean a =
      one ||
      two >> 1 || // one // five
      // two
      // three

      // four
      three;

    boolean a =
      one ||
      two >> 1 || // one
      // two
      // three

      three;

    boolean a =
      one ||
      two >> 1 || // one
      // two
      // three
      three;
  }
}
```

PR #354, opened 2020-01-02
Add option support for trailing comma in enums

clementdessoude:
## What changed with this PR: 
- Add options support for trailing commas in enum and arrays

## Example
**options:**
```sh
--trailing-comma all
```
```java
// Input
public enum Enum {
  ONE, TWO, THREE
}

// Output
public enum Enum {
  ONE, 
  TWO,
  THREE,
}
```

```java
// Input
public class T {
  void t() {
    int[] ints = { 1, 2, 3, };
    int[] ints = { aVeryLongArrayValue, anotherVeryLongArrayValue, andYetAnotherVeryLongArrayValue };
  }
}

// Output
public class T {
  void t() {
    int[] ints = { 1, 2, 3 };
    int[] ints = {
      aVeryLongArrayValue,
      anotherVeryLongArrayValue,
      andYetAnotherVeryLongArrayValue,
    };
  }
}
```

**Relative issues or prs:**
Relative to #313, #314 and #319 

cc @natdempk and @jhaber

### DIFF
--- a/packages/prettier-plugin-java/src/cst-printer.js
+++ b/packages/prettier-plugin-java/src/cst-printer.js
@@ -88,7 +88,10 @@ class CstPrettierPrinter extends BaseJavaCstVisitor {
                   .endOffset
               : node.location.endOffset;
 
-          return this.originalText.substring(startOffset, endOffset + 1);
+          return this.prettierOptions.originalText.substring(
+            startOffset,
+            endOffset + 1
+          );
         } catch (e) {
           throw Error(
             e +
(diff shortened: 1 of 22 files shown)


murdos:
I'm not convinced that this is a feature we should support. 
The prettier philosophy is to be opinionated and have few options, and I don't think a new option should be added here.

clementdessoude:
I agree with you that options should be added very carefully, and this one could create a precedent. However, it is one of the very few options provided by Prettier, so we might consider it anyway

murdos:
Ah, sorry! I was not aware that prettier already provided this option.
So definitely +1, we should indeed support it

jhaber:
I believe array initializers are the only other part of the language that support trailing commas, we may want this option to apply there too:
https://docs.oracle.com/javase/specs/jls/se7/html/jls-10.html#jls-10.6
> A trailing comma may appear after the last expression in an array initializer and is ignored.

For example: `int[] ints = { 1, 2, 3, }`

clementdessoude:
Ok, so let's do this!

I added the trailing comma support in arrays too. I followed Prettier behavior which adds a comma (when the option is enabled) in arrays/objects only if they break

PR #364, opened 2020-01-26
docs: add docs on how to use the plugin

clementdessoude:
## What changed with this PR:

Add docs on how to use the plugin

@pascalgrimaud I would like to have your feedback on this when you have time :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,3 @@
-<p align="center">
-    :construction: Work in Progress! :construction:
-</p>
-
 [![Azure DevOps Build Status][azure-devops-image]][azure-devops-url-main] [![Build Status][travis-image]][travis-url]
 
 [azure-devops-image]: https://dev.azure.com/jhipster/prettier-java/_apis/build/status/jhipster.prettier-java?branchName=master
@@ -12,6 +8,10 @@
 # Prettier Java
 
 ![Prettier Banner](https://raw.githubusercontent.com/prettier/prettier-logo/master/images/prettier-banner-light.png)
+
+## Intro
+
+Prettier is an opinionated code formatter. It enforces a consistent style by parsing your code and re-printing it with its own rules that take the maximum line length into account, wrapping code when necessary.
 
 ## How it works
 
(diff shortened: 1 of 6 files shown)


pascalgrimaud:
yes, thanks. Will have a look tomorrow !

README.md
```diff
@@ -43,44 +43,42 @@ This project contains 2 packages:
...
 
-If the plugin is not automatically loaded:
+### Usage
+
+To reformat all you Java files, run:
 
 ```bash
-# Example where the plugin is locate in node_modules
-prettier --write MyJavaFile.java --plugin=./node_modules/prettier-plugin-java
+prettier --write "**/*.java"
```
pascalgrimaud: it will probably work only if you install prettier globally, right ? Otherwise, you need a specific "scripts" part inside `package.json`

docs/advanced_usage.md
```diff
@@ -0,0 +1,89 @@
...
+To share your Git Hooks, we would suggest to follow this procedure:
+
+1. Create a `package.json` file. For instance:
+   ```bash
+   echo "{}" > package.json
+   ```
+2. Install Husky, Lint-staged, Prettier, the Prettier Java plugin locally `npm install --save-dev husky lint-staged prettier prettier-plugin-java`
+3. Add the following to your `package.json`:
+   ```json
+   "husky": {
```
pascalgrimaud: can it be added inside the `package.json` ?
I need to put this inside `.huskyrc` instead

README.md
```diff
@@ -43,44 +43,42 @@ This project contains 2 packages:
...
-`npm install -g prettier`
-
-or
+- Node version 10+
 
-`yarn global add prettier`
+### Install Prettier and Prettier-Java plugin
 
-### Install plugin
+Since the plugin is meant to be used with Prettier, you need to install it, as well as the prettier-plugin-java package.
```
hdurix: Are you sure? As I understand, `prettier` is a transitive dependency of `prettier-plugin-java` so no need to install both.

See [here](https://github.com/hdurix/nba-stats/blob/master/package.json), I only defined `prettier-plugin-java`.

README.md
```diff
@@ -43,44 +43,42 @@ This project contains 2 packages:
...
+yarn add prettier prettier-plugin-java --dev
+
+# Or globally
+yarn global add prettier prettier-plugin-java
 ```
 
-If the plugin is not automatically loaded:
+### Usage
+
+To reformat all you Java files, run:
```
hdurix: your

docs/advanced_usage.md
```diff
@@ -0,0 +1,89 @@
...
+     "hooks": {
+       "pre-commit": "lint-staged"
+     }
+   },
+   "lint-staged": {
+     "*.{java}": [
```
pascalgrimaud: not sure this syntax works... I need to use something like that: `
`{,src/**/,webpack/}*.{md,json,js,ts,css,scss,java}``

docs/advanced_usage.md
```diff
@@ -0,0 +1,89 @@
...
+       "prettier --write",
```
hdurix: as @pascalgrimaud said, maybe with a NPM script?

docs/advanced_usage.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Prettier plugin
+
+The easiest way to use Prettier-Java in you IDE is to install it locally (`npm install prettier prettier-plugin-java --save-dev`) and install the Prettier extension related to your IDE.
+
+Please refer to the [Prettier Editor integration documentation](https://prettier.io/docs/en/editors.html) for more information.
+
+### FileWatcher plugin
+
+You can also use File Watcher to reformat your code on save. We will describe the procedure to to this in the VSCode and Intellij based IDEs.
```
hdurix: IntelliJ

docs/advanced_usage.md
```diff
@@ -0,0 +1,89 @@
...
+    "isAsync": true,
+    "cmd": "npx prettier --write ${file}",
+    "event": "onFileChange"
+  }
+]
+```
+
+If you have installed the prettier and prettier-plugin-java packages globally, replace `npx prettier --write ${file}` command by `prettier --write ${file}`.
+
+#### Intellij based IDEs:
```
hdurix: IntelliJ

docs/advanced_usage.md
clementdessoude: We could add `npx` before, but is not mandatory, IMO. I tried to setup a clean docker to do some test but didn't manage to setup husky. It works even if I unistall my global dependency to prettier

README.md
clementdessoude: You're right if it is installed locally. When installed globally, the prettier command is not added to the PATH if we only install prettier-plugin-java.

docs/advanced_usage.md
clementdessoude: Thanks ! I need to add a comma before the `java`

docs/advanced_usage.md
clementdessoude: You can do both (at least it works for me)

README.md
clementdessoude: Should be good now :)

PR #380, opened 2020-03-26
Feat/create docs site

### LINKED ISSUE #193: Create Demos/Playground
Examples:
- https://github.com/SAP/ui5-typescript/tree/master/demos/demo-js

Note both these mini example packages include links to gitpod.io workspaces
so no needs to clone and setup an env locally 😄

clementdessoude:
## What changed with this PR:

Add a website at https://prettier-java.tech ! 🚀

Screen Shot 2020-03-26 at 11 20 36
Screen Shot 2020-03-26 at 11 20 50

## Relative issues or prs:

Close #193


noahbetzen-wk:
The playground page works for me on Chrome but not Firefox/Safari:
Screen Shot 2020-03-26 at 11 34 47 AM

I don't think this should hold up the PR for now, but perhaps a disclaimer on the page saying "This page will only render on Chrome for now" would be useful.

website/blog/2020-01-19-v0.7.0.md
```diff
@@ -0,0 +1,328 @@
...
+
+  ```java
+  // Input
+  public enum Enum {
+    ONE,
+    TWO,
+    THREE
+  }
+
+  // Output
```
imod: I don't see a difference between the input and output

imod:
can you please explain a typical workflow? how would you use it?
I understand IDEA does have a prettier plugin and to ensure consistent formatting one can add a Git pre-commit hook.
But how do you use it with e.g. Eclipse? As far as I know, there is no prettier plugin for Eclipse. How would you use it in this case? That would mean when ever a user uses the eclipse formatter, the whole class already changes all over and one has a hard time to compare the local changes in git until prettier was triggered (manually). 

Maybe there is a way to generate a eclipse formatter configuration which comes as close as possible to the prettier format?

PR #305, opened 2019-12-01
feature: improve throws rendering

clementdessoude:
Proposition to fix #286

I'm not sure about how to format some cases so feedbacks are welcome 

(cc @murdos as you are the one who created the issue)

Issue #286, opened 2019-10-20
Improve formatting of method will lot of parameters and with exception


murdos:
Source:
```
    public LoggingConfiguration(@Value("${spring.application.name}") String appName,
        @Value("${server.port}") String serverPort,
        JHipsterProperties jHipsterProperties,
        ObjectMapper mapper)
        throws JsonProcessingException {
        // ...
    }
```

Formatted with prettier 0.4.0:
```
    public LoggingConfiguration(
        @Value("${spring.application.name}") String appName,
        @Value("${server.port}") String serverPort,
        JHipsterProperties jHipsterProperties,
        ObjectMapper mapper
    )
        throws JsonProcessingException {
        // ...
    }
```

I'm not really sure what the result should be, but the position of `)` after the last parameter is a bit strange in regards to the exception.
Maybe:
```
    public LoggingConfiguration(
        @Value("${spring.application.name}") String appName,
        @Value("${server.port}") String serverPort,
        JHipsterProperties jHipsterProperties,
        ObjectMapper mapper
    ) throws JsonProcessingException {
        // ...
    }
```

clement26695:
I agree with you. I am struggling with one case: how would you format the following ?
```java
public void test() throws Exception1, Exception2, Exception3, Exception4, Exception5, Exception6 {}
```

Option 1:
```java
public void test() 
    throws Exception1, Exception2, Exception3, Exception4, Exception5, Exception6 {}
```
Option 2:
```java
public void test() throws 
    Exception1, Exception2, Exception3, Exception4, Exception5, Exception6 {}
```
Option 3:
```java
public void test() throws Exception1, 
    Exception2, 
    Exception3, 
    Exception4, 
    Exception5, 
    Exception6 {}
```
Option 4:
```java
public void test() throws 
    Exception1, 
    Exception2, 
    Exception3, 
    Exception4, 
    Exception5, 
    Exception6 {}
```
Option 5:
```java
public void test() throws Exception1, Exception2, Exception3, Exception4, Exception5, Exception6 {}
```

clement26695:
Does this formatting looks good ?
```java
class T {
  public void test(String one, String one, String one, String one, String one, String one) {

  }

  public void test(String one, String one, String one, String one, String one, String one) throws Exception, Exception, Exception {

  }

  public void test(String one, String one, String one, String one, String one, String one) throws Exception {

  }

  public void test(String one, String one) throws Exception, Exception, Exception, Exception {

  }

[...]

Issue #271, opened 2019-09-25
Blank line removed after class declaration


murdos:
This might be done on purpose, but I'm opening this issue to be able to discuss it.
Before:
```
public class JWTFilter extends GenericFilterBean {

    public static final String AUTHORIZATION_HEADER = "Authorization";

    private TokenProvider tokenProvider;
```

After:
```
public class JWTFilter extends GenericFilterBean {
    public static final String AUTHORIZATION_HEADER = "Authorization";

    private TokenProvider tokenProvider;
```

I think it would be more readable to add/keep the extra line.

In the following example it's hard to distinguish between the end of the class declaration (and its superclass and interfaces) and the first field:
```
public class MyLoooooongClass
    extends AbstractAuditingEntity
    implements Serializable, MyInterface<AAAAAAA, BBBBBBBBBBB> {
    private static final long serialVersionUID = 1L;

    @Id
    @GeneratedValue(strategy = GenerationType.SEQUENCE)
```

Also if the first "content" of the class is an inner class the blank line is added:
```
 /**
 * some javadoc
 */
public class MyEntityCriteria implements Serializable, Criteria {

    /**
     * Class for filtering MyEntity
     */
    public static class MyEntityFilter extends Filter<MyEntity> {

        public MyEntityFilter() {}
```

clement26695:
@murdos Do you think we should always add a blank line at the beginning of class bodies and interfaces bodies ?

murdos:
Yes, that's what I would like (when there's a non empty body of course).
But I'm not sure there's a consensus.

pascalgrimaud:
Adding a bounty on this

benjavalero:
> @murdos Do you think we should always add a blank line at the beginning of class bodies and interfaces bodies ?

What about the end of a class? It usually contains a method, so after formatting there is no blank line between the ending `}` of the method and the ending `}`of the class.

jhaber:
> > @murdos Do you think we should always add a blank line at the beginning of class bodies and interfaces bodies ?
> 
> What about the end of a class? It usually contains a method, so after formatting there is no blank line between the ending `}` of the method and the ending `}`of the class.

I don't think adding a newline between the last method and end of class is a common style so I would advocate for prettier not doing it. 

[...]

PR #384, opened 2020-04-10
fix: throws rendering

jhaber:
This is a modified version of #305 that makes a more minor change. Based on #286 it seems like there's not a clear answer on how formatting should work when methods throw lots of exceptions. However, the current behavior where there's an extra newline is basically a bug, for example:

```java
public void method(
  Arg arg1,
  Arg arg2
)
  throws Exception {}
```

This PR attempts to fix the bug, while punting on the question of a more complete behavior change.

The tests should give a good idea of the behavior I was going for. Unfortunately, my prettier fu is pretty weak so a few of the tests are failing and I'm having a hard time getting things just right

### DIFF
--- a/packages/prettier-plugin-java/src/printers/classes.js
+++ b/packages/prettier-plugin-java/src/printers/classes.js
@@ -1,6 +1,6 @@
 "use strict";
 const _ = require("lodash");
-const { line, softline, hardline } = require("prettier").doc.builders;
+const { ifBreak, line, softline, hardline } = require("prettier").doc.builders;
 const {
   getBlankLinesSeparator,
   reject,
@@ -402,11 +402,6 @@ class ClassesPrettierVisitor {
     const declarator = this.visit(ctx.methodDeclarator);
     const throws = this.visit(ctx.throws);
 
-    let throwsPart = "";
-    if (throws) {
-      throwsPart = indent(rejectAndConcat([softline, throws]));
-    }
-
     return group(
       concat([
         rejectAndJoin(" ", [
@@ -414,7 +409,7 @@ class ClassesPrettierVisitor {
           rejectAndJoin(line, annotations),
           result,
           declarator,
-          throwsPart
+          throws
         ])
       ])
     );
@@ -514,7 +509,13 @@ class ClassesPrettierVisitor {
 
   throws(ctx) {
     const exceptionTypeList = this.visit(ctx.exceptionTypeList);
-    return join(" ", [ctx.Throws[0], exceptionTypeList]);
+    const throwsDeclaration = join(" ", [ctx.Throws[0], exceptionTypeList]);
+    return group(
+      ifBreak(
+        indent(rejectAndConcat([softline, throwsDeclaration])),
+        throwsDeclaration
+      )
+    );
   }
 
   exceptionTypeList(ctx) {
@@ -554,11 +555,6 @@ class ClassesPrettierVisitor {
     const throws = this.visit(ctx.throws);
     const constructorBody = this.visit(ctx.constructorBody);
 
-    let throwsPart = "";
-    if (throws) {
-      throwsPart = indent(rejectAndConcat([softline, throws]));
-    }
-
     return rejectAndJoin(" ", [
       group(
         rejectAndJoin(hardline, [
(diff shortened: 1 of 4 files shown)


jhaber:
Updated test output to match current behavior so the tests pass, here's the diff:
https://github.com/HubSpot/prettier-java/commit/6d41b672ed29080800020d4a156deef4be7a1188

There's still an extra newline but only when the throws clause is really long

packages/prettier-plugin-java/src/printers/classes.js
```diff
@@ -511,7 +506,8 @@ class ClassesPrettierVisitor {
 
   throws(ctx) {
     const exceptionTypeList = this.visit(ctx.exceptionTypeList);
-    return join(" ", [ctx.Throws[0], exceptionTypeList]);
+    const throwsDeclaration = join(" ", [ctx.Throws[0], exceptionTypeList]);
+    return group(indent(rejectAndConcat([softline, throwsDeclaration])));
```
clementdessoude, line 510: We could add an extra indent here, so that the throws + exceptionTypeList is not at the same level of the rest of the body, WDYT ?

clementdessoude:
Thank you for this @jhaber! One minor comment and we can merge it :)

packages/prettier-plugin-java/src/printers/classes.js
jhaber: Hmm, would that be the only place in prettier-java where a double-indent is used? There is a similar issue with a really long class declaration, for example:

```java
public class MyReallyLongClassName
  extends SomeOtherReallyLongClass
  implements SomeReallyLongInterface {
  private Field field;

  public Field someMethodWithAReallyLongName()
    throws IOException {
    return field;
  }
}
```

I wonder if the fix in both cases could be to force a newline if the class signature or throws clause linebreaks? This would look like:

```java
public class MyReallyLongClassName
  extends SomeOtherReallyLongClass
  implements SomeReallyLongInterface {

  private Field field;

  public Field someMethodWithAReallyLongName()
    throws IOException {

    return field;
  }
}
```

jhaber:
Thanks! There is still still an extraneous newline when the throws clause is really long, for example:
https://github.com/HubSpot/prettier-java/commit/6d41b672ed29080800020d4a156deef4be7a1188

I have a hunch I need to use `ifBreak` somewhere, but couldn't figure out where

packages/prettier-plugin-java/src/printers/classes.js
clementdessoude: There are other places where we could add some double indents indeed.

It might indeed be better to add an extra blank line, but it is more difficult to implement, IMO, as the formatting of the class body is not in the same node as the inheritance/interface declaration.

jhaber: Would it be easier if class bodies always started with an empty line? (basically #271) 

I guess that wouldn't help the method case though, since you wouldn't want methods to unconditionally start with an empty line

jhaber: Also getting back to the original suggestion, double indent (or continuation indent) is pretty common in Java formatters but doesn't seem to be idiomatic in prettier as far as I can tell (this issue didn't seem to go anywhere https://github.com/prettier/prettier/issues/5865). 

If we introduced the idea of continuation indent, I think we'd want to apply it consistently. Off the top of my this would include class signatures:
```java
public class MyReallyLongClassName
  ->extends SomeOtherReallyLongClass
  ->implements SomeReallyLongInterface {
```

Method signatures:
```java
public void myMethod(
  ->String argOne
  ->String argTwo
  ->String argThree
) {}
```

Method call chains:
```java
return someCollectionType
  ->.stream()
  ->.filter(x -> x != null)
  ->.collect(Collectors.toList());
```

And probably a bunch of other statements as well. This is definitely one way to go, but would be a pretty massive formatting change at this point. Might be better to move discussion to an issue if that's under consideration

jhaber:
Any more thoughts on this PR? We've been formatting all of our code with this change since April and haven't noticed any undesirable side effects.

jhaber:
👋 bump on this PR. We're continuing to apply it on top of each prettier-java release for our internal use, but it is a bit of a maintenance burden

clementdessoude:
I'll think I merge it next week. I will be on holidays so I'll finally have some time :)

pascalgrimaud:
@Shaolans @Hawkurane : can one of you have a quick look and merge this plz ?

clementdessoude:
Sorry it took so long @jhaber. I'm merging this. Thank you a lot !

PR #385, opened 2020-04-13
fix: empty enums

jhaber:
## What changed with this PR:

Fix handling of empty enums to not insert a trailing comma

## Example

The tests show examples of the new behavior. Without this PR, the output is:

```java
public enum EmptyEnum {
  ,
}

public enum EmptyEnumWithComment {
  ,
  // comment
}
```

Which doesn't compile

### DIFF
--- a/packages/prettier-plugin-java/src/printers/classes.js
+++ b/packages/prettier-plugin-java/src/printers/classes.js
@@ -674,12 +674,14 @@ class ClassesPrettierVisitor {
     const enumConstantList = this.visit(ctx.enumConstantList);
     const enumBodyDeclarations = this.visit(ctx.enumBodyDeclarations);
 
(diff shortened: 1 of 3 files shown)


packages/prettier-plugin-java/src/printers/classes.js
```diff
@@ -674,12 +674,14 @@ class ClassesPrettierVisitor {
     const enumConstantList = this.visit(ctx.enumConstantList);
     const enumBodyDeclarations = this.visit(ctx.enumBodyDeclarations);
 
+    const hasEnumConstants = ctx.enumConstantList !== undefined;
```
jhaber, line 677: not sure if this is the best way to implement this check

Shaolans:
LGTM

PR #399, opened 2020-05-04
feat: upload binary artifacts on github release

### LINKED ISSUE #326: Create a standalone binary artifact
I would be awesome to provide a standalone artifact to run Prettier for Java in an ecosystem without node/npm

See #317

Shaolans:
## What changed with this PR:
This PR will provide binary artifacts on github release.
Working on linux but I have tested for macos or windows thought.

Close #326

## Example

```java
// Input

// Output
```

### DIFF
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -6,7 +6,7 @@ jobs:
   generate:
     env:
       # The version needs to match with the one used in the printer
-      PRETTIER_API: 2.0.0
+      PRETTIER_CORE_LIB_VERSION: 2.0.0
     name: Create release artifacts
     runs-on: ubuntu-latest
     steps:
@@ -34,7 +34,7 @@ jobs:
         uses: actions/checkout@v2
         with:
           repository: prettier/prettier
-          ref: refs/tags/${{ env.PRETTIER_API }}
+          ref: refs/tags/${{ env.PRETTIER_CORE_LIB_VERSION }}
           path: prettier
       - name: Link Prettier-java to Prettier
         run: |
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -37,6 +37,7 @@ jobs:
           ref: refs/tags/${{ env.PRETTIER_CORE_LIB_VERSION }}
           path: prettier
       - name: Link Prettier-java to Prettier
+        # We need to give the path to the plugin for pkg
         run: |
           cd $GITHUB_WORKSPACE/prettier
           npm i
--- a/.github/workflows/binaries.yml
+++ b/.github/workflows/binaries.yml
@@ -1,0 +1,55 @@
+on:
+  release:
+    types: [created]
+name: Binary artifacts
+jobs:
+  generate:
+    env:
+      # The version needs to match with the one used in the printer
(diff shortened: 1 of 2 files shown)


.github/workflows/binaries.yml
```diff
@@ -0,0 +1,55 @@
+on:
+  release:
+    types: [created]
+name: Binary artifacts
+jobs:
+  generate:
+    env:
+      # The version needs to match with the one used in the printer
+      PRETTIER_API: 2.0.0
```
clementdessoude, line 9: I'm not convinced by the API terminology. Can we use `PRETTIER_CORE_LIB_VERSION` instead or something more explicit

.github/workflows/binaries.yml
```diff
@@ -0,0 +1,55 @@
...
+          repository: prettier/prettier
+          ref: refs/tags/${{ env.PRETTIER_API }}
+          path: prettier
+      - name: Link Prettier-java to Prettier
+        run: |
+          cd $GITHUB_WORKSPACE/prettier
+          npm i
+          npm link prettier-plugin-java
+          mv package.json package.json.old
+          jq '.dependencies["prettier-plugin-java"] = "./node_modules/prettier-plugin-java"' package.json.old | cat > package.json
```
clementdessoude, line 45: I do not understand this line: is the `npm link prettier-plugin-java` command not sufficient ? If not, can you add a comment to add some explanation ?

Shaolans: It looks like pkg is solely relying on the package.json and does not seems to use node_modules folder. I had to manually route the dependency to the right one.
However this is just my two cents guess as I only did some tests and don't really know how pkg works from the inside.

Shaolans:
> Is it good to go or does it need some further work ?

It should be good to go, but I only tested the linux artifact.

clementdessoude:
Thanks @Shaolans ! If you find a way to reduce the binary size, it would be wonderful ! I just need to create the changelog and will do the release

Shaolans:
Yes, it is quite a bummer that this weight 100mb+ I was quite surprised

jhaber:
I'm guessing that's mainly coming from Node.js being bundled? And I would imagine there's at least 3 versions included (Linux/OSX/Windows)

I bundled prettier-java into a single JS file using webpack which only came out ~3MB if I remember correctly, but this requires the user to bring their own Node.js

Shaolans:
I think the Node.js bundle is about ~40MB.
I found a way to compress the binary with gzexe for the linux binary but at a great cost. It quadruple the runtime time on my machine...

clementdessoude:
@Shaolans The hook is failing on Mac Os :/

Shaolans:
I think gnu-sed is mandatory, have you this installed ?

Walnussbaer:
Do you still publish these binaries somewhere? I can't find them and you docs page only mentions installation via npm/yarn/pnpm.

Walnussbaer:
> @Walnussbaer It appears that the GitHub Actions workflow for these binaries has been broken for quite some time. If you require or are interested in using these binaries, please create a new issue on this repository describing your intended use case and I can look into fixing the workflow.

Thanks for the reply. We have to evalutate our current situation. I will create a ticket if needed.

PR #407, opened 2020-06-05
Add more context to the Maven plugin usage

shepherdjerred:
## What changed with this PR:
Improving the README documentation so that it shows what part of the pom.xml the plugin entry belongs in. It's been a while since I've used Maven, so it took me a little while to figure out how to actually use this plugin in a fresh pom.xml without a <plugins> block.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ To see an advanced usage: please go to the [Advanced Usage](./docs/advanced_usag
 ## Maven plugin
 
 A neat maven plugin for prettier-java was made by developers from HubSpot. \
-It is currently available at:
+Add it to the `plugins` section of your `build` configuration
 
 ```xml
 <build>
@@ -95,7 +95,7 @@ It is currently available at:
     <plugin>
       <groupId>com.hubspot.maven.plugins</groupId>
         <artifactId>prettier-maven-plugin</artifactId>
-        <!-- Find the latest version -->
+        <!-- Find the latest version at https://github.com/jhipster/prettier-java/releases -->
         <version>0.8</version>
     </plugin>
   </plugins>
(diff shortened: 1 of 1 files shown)


Shaolans:
Hi, thank you for your contribution !
I added the code snippet in order to indicate where the plugin is available at, that is why it is also recommended to checkout the [project](https://github.com/HubSpot/prettier-maven-plugin) if you want to use it.
As you had troubled with using the maven plugin, maybe our doc need indeed some improvements.

README.md
```diff
@@ -90,12 +90,16 @@ A neat maven plugin for prettier-java was made by developers from HubSpot. \
 It is currently available at:
```
Shaolans, line 90: If we no longer point where the plugin is available at but rather the minimal configuration we should change this text accordingly.

README.md
```diff
@@ -90,12 +90,16 @@ A neat maven plugin for prettier-java was made by developers from HubSpot. \
...
+      <groupId>com.hubspot.maven.plugins</groupId>
+        <artifactId>prettier-maven-plugin</artifactId>
+        <!-- Find the latest version -->
+        <version>0.8</version>
+    </plugin>
+  </plugins>
+</build>
 ```
 
 If you would like to use this plugin, we recommend you to check their [project](https://github.com/HubSpot/prettier-maven-plugin) as is it well documented.
```
Shaolans, line 105: Here as well

README.md
shepherdjerred: I've updated the text to address this comment. Let me know what you think.

README.md
shepherdjerred: What do you think should be changed here? To me it is clear as it stands.

Shaolans: You are right, nevermind :)

README.md
Shaolans: LGTM !

PR #432, opened 2020-11-07
Feat: support new switch rules

### LINKED ISSUE #277: Support for Java 13
Support for Java 13 and language previews such as [JEP 355: Text Blocks](https://openjdk.java.net/jeps/355) would be awesome.

Thoughts?

clementdessoude:
## What changed with this PR:

Support new switch rules and yield statements:

## Example

```java

class Test {
  
  // switch rules
  static void howManyAgain(int k) {
    switch (k) {
      case 1 -> System.out.println("one");
      case 2 -> {
        System.out.println("two");
      }
      case 3 -> throw new Exception(e);
    }
  }
  
  // yield statements
  public int calculate(Day d) {
    switch (d) {
      case SATURDAY ,SUNDAY -> d.ordinal();
      default -> {
        int len = d.toString().length();
        yield len * len;
      }
    }
    return;
  }

  public int calculate(Day d) {
    return switch (d) {
      case SATURDAY ,SUNDAY -> d.ordinal();
      default -> {
        int len = d.toString().length();
        yield len * len;
      }
    };
  }
}

```

## Relative issues or prs:

Closes #277 😱


jhaber:
🙇 just wanted to say thanks for continuing to update the parser and formatter for new language constructs

clementdessoude:
Haha, thanks @jhaber, that is actually the fun part :)

PR #476, opened 2021-06-04
feat: support instance of pattern matching

### LINKED ISSUE #434: Support preview feature: Pattern Matching for instanceof
# Why ?

Patterns instance of are a preview feature from Java 14 & 15. We want to support it because we are hipsters ! 🤘

## Resources:

- https://docs.oracle.com/javase/specs/jls/se15/preview/specs/patterns-instanceof-jls.html
- https://docs.oracle.com/en/java/javase/15/language/pattern-matching-instanceof-operator.html

clementdessoude:
## What changed with this PR:

Support instance of pattern matching: https://docs.oracle.com/javase/specs/jls/se16/html/jls-15.html#jls-InstanceofExpression

## Example

```java
// Input
class T {
    static String formatter(Object o) {
        String formatted = "unknown";
        if (o instanceof Integer i || p instanceof Point || q instanceof Circle c || r instanceof Square) {
            formatted = String.format("int %d", i);
        } else if (o instanceof Long l) {
            formatted = String.format("long %d", l);
        } else if (o instanceof Double d) {
            formatted = String.format("double %f", d);
        } else if (o instanceof String s) {
            formatted = String.format("String %s", s);
        }
        return formatted;
    }
}

// Output
class T {

  static String formatter(Object o) {
    String formatted = "unknown";
    if (
      o instanceof Integer i ||
      p instanceof Point ||
      q instanceof Circle c ||
      r instanceof Square
    ) {
      formatted = String.format("int %d", i);
    } else if (o instanceof Long l) {
      formatted = String.format("long %d", l);
    } else if (o instanceof Double d) {
      formatted = String.format("double %f", d);
    } else if (o instanceof String s) {
      formatted = String.format("String %s", s);
    }
    return formatted;
  }
}

```

## Relative issues or prs:

Closes #434

### DIFF
--- a/packages/java-parser/src/productions/expressions.js
+++ b/packages/java-parser/src/productions/expressions.js
@@ -123,7 +123,15 @@ function defineRules($, t) {
           {
             ALT: () => {
               $.CONSUME(t.Instanceof);
-              $.SUBRULE($.referenceType);
+              $.OR1([
+                {
+                  GATE: () => this.BACKTRACK_LOOKAHEAD($.pattern),
+                  ALT: () => $.SUBRULE($.pattern)
+                },
+                {
+                  ALT: () => $.SUBRULE($.referenceType)
+                }
+              ]);
             }
           },
           {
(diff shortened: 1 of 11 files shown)


packages/java-parser/src/productions/expressions.js
```diff
@@ -123,7 +123,15 @@ function defineRules($, t) {
           {
             ALT: () => {
               $.CONSUME(t.Instanceof);
-              $.SUBRULE($.referenceType);
+              $.OR1([
```
Shaolans, line 126: just ```OR``` is not enough ?

Shaolans (approved):

PR #558, opened 2022-10-22
Update to prettier 3

clementdessoude:
## What changed with this PR:

Update prettier dependency to v3

## Relative issues or prs:

#557

### DIFF
--- a/packages/prettier-plugin-java/src/printers/prettier-builder.ts
+++ b/packages/prettier-plugin-java/src/printers/prettier-builder.ts
@@ -1,12 +1,9 @@
 "use strict";
 import { IToken } from "java-parser";
-import { doc } from "prettier";
-import Doc = doc.builders.Doc;
-
-const prettier = require("prettier").doc.builders;
+import { builders } from "prettier/doc";
+import Doc = builders.Doc;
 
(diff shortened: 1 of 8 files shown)


fisker:
You may want remove `@types/prettier`

clementdessoude:
> You may want remove `@types/prettier`

Yes, indeed ! Thanks for pointing this out @fisker ! I think everything is working fine with prettier v3 here.

I will keep this issue opened until the final prettier v3.0.0 is released, and update it with intermediate version if there are any

packages/prettier-plugin-java/src/printers/prettier-builder.ts
```diff
@@ -2,8 +2,6 @@
 import { IToken } from "java-parser";
 import { doc } from "prettier";
 import Doc = doc.builders.Doc;
-import Concat = doc.builders.Concat;
-import Fill = doc.builders.Fill;
 
 const prettier = require("prettier").doc.builders;
```
fisker, line 6: I think this  require() can be removed too. `doc` is imported, and it's not type only.

clementdessoude:
Oh, yeah! 🥳
I'll try to update this weekend!

PR #588, opened 2023-07-21
ci: downgrade E2E JHipster samples

jtkiesel:
## What changed with this PR:

All of the JHipster repositories that this project uses as samples for its E2E tests had their main branches updated last week with 2 breaking changes:

1. They now require Java 17 to build.
2. Their `mvnw` script is no longer executable (@pascalgrimaud apologies for the notification, but I thought JHipster might want to be made aware of this issue, as I assume this was changed by mistake).

Both of these changes broke this repository's CI checks. The first change would be easy enough for us to resolve, we can simply upgrade the version of Java in `github-ci.yml`. However, the second change is not as easy for us to fix, nor should we have to fix it here, so we'll likely need to wait until those other repositories are updated to re-set the executable bit on their `mvnw` script.

To work around this issue while we wait, this PR downgrades the E2E JHipster samples to the latest working tag.

### DIFF
--- a/packages/prettier-plugin-java/scripts/clone-samples.js
+++ b/packages/prettier-plugin-java/scripts/clone-samples.js
@@ -9,7 +9,7 @@ const samplesDir = path.resolve(__dirname, "../samples");
 const core = [
   {
     repoUrl: "https://github.com/jhipster/jhipster-sample-app",
-    branch: "main"
+    branch: "v7.9.3"
   },
   {
     repoUrl: "https://github.com/jhipster/jhipster",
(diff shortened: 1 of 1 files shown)


pascalgrimaud:
No worry for notification @jtkiesel :)

I didn't follow closely what was done in generator-jhipster, since I'm focusing on jhipster-lite for my personal and professional uses.
Since it looks not ready, I think it should be merged to not block contributions / fixes

cc @DanielFran @clementdessoude : I'm going to merge this, waiting some fix. Is it fine for you?

PR #591, opened 2023-07-29
fix: improve throws formatting

### LINKED ISSUE #583: Use different indentation level for `throws`
I tried Prettier-java today and saw this formatting result:

```java
public void myMethodWithLongName(final boolean myArg1, ...)
    throws IOException {
    final String myVariable = "123";
    ...
```

This makes the `throws` line and the 1st line of the method body visually indistinguishable.  Is it possible to change this behavior?

I think a potential solution would be: to allow a different indentation level (such as 8) for line break indentations.

jtkiesel:
## What changed with this PR:

Constructors/methods with `throws` clauses now break on their parameters before breaking on `throws`, and the comma-separated list of exception classes will now break rather than violating printWidth. This was loosely designed based on the way Prettier JavaScript formats TypeScript methods with return types. Open to feedback on the formatting I chose to go with.

### Input

```java
class Example {

  Example(String arg1, String arg2, String arg3, String arg4)
    throws RuntimeException, RuntimeException, RuntimeException {
    throw new RuntimeException();
  }

  Example(String arg1)
    throws RuntimeException, RuntimeException, RuntimeException, RuntimeException {
    throw new RuntimeException();
  }

  Example()
    throws RuntimeException, RuntimeException, RuntimeException, RuntimeException {
    throw new RuntimeException();
  }

  void example1(String arg1, String arg2, String arg3, String arg4)
    throws RuntimeException, RuntimeException, RuntimeException {
    throw new RuntimeException();
  }

  void example2(String arg1)
    throws RuntimeException, RuntimeException, RuntimeException, RuntimeException {
    throw new RuntimeException();
  }

  void example3()
    throws RuntimeException, RuntimeException, RuntimeException, RuntimeException {
    throw new RuntimeException();
  }
}
```

### Output

```java
class Example {

  Example(
    String arg1,
    String arg2,
    String arg3,
    String arg4
  ) throws RuntimeException, RuntimeException, RuntimeException {
    throw new RuntimeException();
  }

  Example(
    String arg1
  ) throws RuntimeException,
    RuntimeException,
    RuntimeException,
    RuntimeException {
    throw new RuntimeException();
  }

  Example()
    throws RuntimeException,
    RuntimeException,
    RuntimeException,
    RuntimeException {
    throw new RuntimeException();
  }

  void example1(
    String arg1,
    String arg2,
    String arg3,
    String arg4
  ) throws RuntimeException, RuntimeException, RuntimeException {
    throw new RuntimeException();
  }

  void example2(
    String arg1
  ) throws RuntimeException,
    RuntimeException,
    RuntimeException,
    RuntimeException {
    throw new RuntimeException();
  }

  void example3()
    throws RuntimeException,
    RuntimeException,
    RuntimeException,
    RuntimeException {
    throw new RuntimeException();
  }
}
```

## Relative issues or prs:

Might be enough to close #583? I specifically did **not** use a continuation indent as was mentioned in the issue as a potential solution, as this would be the first of such an indent in Prettier Java. Because of that, not *all* cases of constructors/methods with `throws` clauses are easier to read, but in my opinion the example given in that ticket (a method with at least 1 parameter and a short `throws` clause) is easier to read, since the parameters are now broken first.


jtkiesel:
@clementdessoude I'm curious what your thoughts are on the formatting decisions I made here. Could you take a look, whenever you have the time?

packages/prettier-plugin-java/src/printers/classes.ts
```diff
@@ -652,15 +658,16 @@ export class ClassesPrettierVisitor extends BaseCstPrettierPrinter {
     return printTokenWithComments(this.getSingle(ctx) as IToken);
   }
 
-  throws(ctx: ThrowsCtx) {
+  throws(ctx: ThrowsCtx, opts?: { throwsGroupId?: symbol }) {
```
clementdessoude, line 661: Should throwsGroupId really be optional, or should we force it to be defined so we are "forced" to align throws formating for constructor/interface methods, etc. as well ?

jtkiesel: Good point! I totally forgot to implement this for interface methods. I went ahead and made this required as you suggested, and implemented the same throws formatting for interface methods.

PR #595, opened 2023-08-06
fix: stable format for if expression with trailing comment

### LINKED ISSUE #592: Formatting not stable with comment in if statement
**Prettier-Java 2.2.0**

```java
$ yarn add prettier prettier-plugin-java
$ cat initial.java
package main;

class Main {

    if (true) // comment
    {
      System.out.println("Oops");
    }
  }
}
package main;

class Main {

    if (
      true
    ) { // comment
      System.out.println("Oops");
    }
  }
}
package main;

class Main {

    if (true) { // comment
      System.out.println("Oops");
    }
  }
}
```

**Expected behavior:**

The prettier-java-plugin should only require one format execution to properly format the code.

jtkiesel:
## What changed with this PR:

If statements whose expressions have trailing comments are now formatted stably, and if statements in general are formatted more closely to the way that Prettier JavaScript does.

### Input

```java
class Example {

  void example() {
    if (true) // comment
      System.out.println("Oops");

    if (true) {
      // comment
    }

    if (true) // comment
    {}

    if (true) // comment
    {
      System.out.println("Oops");
    }

    if (true) // comment
    {
      if (true) {}
    }
  }
}
```

### Output

```java
class Example {

  void example() {
    if (true)
      // comment
      System.out.println("Oops");

    if (true) {
      // comment
    }

    if (true) {
      // comment
    }

    if (true) {
      // comment
      System.out.println("Oops");
    }

    if (true) {
      // comment
      if (true) {}
    }
  }
}
```

## Relative issues or prs:

Closes #592


packages/prettier-plugin-java/src/printers/blocks-and-statements.ts
```diff
@@ -192,12 +193,19 @@ export class BlocksAndStatementPrettierVisitor extends BaseCstPrettierPrinter {
   }
 
   ifStatement(ctx: IfStatementCtx) {
+    handleCommentsIfStatement(ctx);
```
clementdessoude, line 196: Is this issue only happening on if statements ?

jtkiesel: Looks like it also happens with for loops, while loops, and switch statements. I'll see about fixing those cases as well.

jtkiesel: Latest push now handles for/while/do loops and switch/catch statements as well.

clementdessoude:
Thank you, @jtkiesel ! It seems that quite a lot of logic is added in order to handle this edge case. I would like to dig a bit more into the issue to see if we can simplify some things before merging this. My objective would be to look at it by the end of the week (and the other pending PR) so we could release a new version of Prettier this weekend :)

clementdessoude:
I looked at it more closely yesterday, but was not able to wrap it up, sorry. I propose to make a new release with the changes that were made since 2.2.0 (https://github.com/jhipster/prettier-java/pull/600), and finish this afterwards.

koppor:
@clementdessoude @jtkiesel Did you find time to look into this? 😇

jtkiesel:
Superseded by #731.

PR #606, opened 2023-09-17
fix: separate binary expression from pattern

### LINKED ISSUE #605: Line break between instanceof and final
**Prettier-Java 2.3.1**

```sh
# Options (if any):
--print-width 80
```

**Input:**

```java
  public boolean test(final Object obj) {
    return obj instanceof final Integer x && (x == 5 || x == 6 || x == 7 || x == 8 || x == 9 || x == 10 || x == 11);
  }
```

**Output:**

```java
  public boolean test(final Object obj) {
    return (
      obj instanceof
      final Integer x &&
      (x == 5 || x == 6 || x == 7 || x == 8 || x == 9 || x == 10 || x == 11)
    );
  }
```

**Expected behavior:**
```java
  public boolean test(final Object obj) {
    return (
      obj instanceof final Integer x &&
      (x == 5 || x == 6 || x == 7 || x == 8 || x == 9 || x == 10 || x == 11)
    );
  }
```

jtkiesel:
## What changed with this PR:

Rather than treating the optional binary expression following a pattern as part of the pattern itself, treat the entire expression as a binary expression, of which the pattern is a part, so that the usual binary expression formatting applies.

### Input

```java
class Example {

  public boolean test(final Object obj) {
    return obj instanceof final Integer x && (x == 5 || x == 6 || x == 7 || x == 8 || x == 9 || x == 10 || x == 11);
  }
}
```

### Output

```java
class Example {

  public boolean test(final Object obj) {
    return (
      obj instanceof final Integer x &&
      (x == 5 || x == 6 || x == 7 || x == 8 || x == 9 || x == 10 || x == 11)
    );
  }
}
```

## Relative issues or prs:

Closes #605

### DIFF
--- a/packages/java-parser/api.d.ts
+++ b/packages/java-parser/api.d.ts
@@ -3498,8 +3498,6 @@ export interface PatternCstNode extends CstNode {
 
 export type PatternCtx = {
   primaryPattern: PrimaryPatternCstNode[];
-  AndAnd?: IToken[];
-  binaryExpression?: BinaryExpressionCstNode[];
 };
 
 export interface PrimaryPatternCstNode extends CstNode {
--- a/packages/java-parser/src/productions/blocks-and-statements.js
+++ b/packages/java-parser/src/productions/blocks-and-statements.js
@@ -245,10 +245,6 @@ function defineRules($, t) {
       { ALT: () => $.CONSUME(t.Null) },
       { ALT: () => $.CONSUME(t.Default) },
       {
-        GATE: () => this.BACKTRACK_LOOKAHEAD($.pattern),
-        ALT: () => $.SUBRULE($.pattern)
-      },
-      {
         GATE: () => tokenMatcher($.LA(1).tokenType, t.Null) === false,
         ALT: () => $.SUBRULE($.caseConstant)
       }
--- a/packages/java-parser/src/productions/expressions.js
+++ b/packages/java-parser/src/productions/expressions.js
@@ -116,17 +116,23 @@ function defineRules($, t) {
   });
 
   $.RULE("binaryExpression", () => {
-    $.SUBRULE($.unaryExpression);
+    $.OR([
+      {
+        GATE: () => this.BACKTRACK_LOOKAHEAD($.pattern),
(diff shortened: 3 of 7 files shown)


packages/java-parser/api.d.ts
```diff
@@ -3498,8 +3498,6 @@ export interface PatternCstNode extends CstNode {
 
 export type PatternCtx = {
   primaryPattern: PrimaryPatternCstNode[];
-  AndAnd?: IToken[];
```
clementdessoude, line 3501: I am not a fan of modifying the CST in order to make it easier to print the code afterwards, as it makes the java parser maintenance much more difficult for each deviation from the spec

jtkiesel: @clementdessoude I looked into the spec while implementing this solution, because I had similar concerns. What I found was that a [Pattern](https://docs.oracle.com/javase/specs/jls/se20/html/jls-14.html#jls-Pattern) is a [TypePattern](https://docs.oracle.com/javase/specs/jls/se20/html/jls-14.html#jls-TypePattern), which is a [LocalVariableDeclaration](https://docs.oracle.com/javase/specs/jls/se20/html/jls-14.html#jls-LocalVariableDeclaration), meaning that according to the spec, a Pattern should not contain AndAnd or a binaryExpression within it. I made this change because it seemed to make the parser more aligned with the spec.

All that being said, this solution is failing tests because it seems to have slowed down parsing, so I'll need to find a fix for that. But I do think this change to the CST is appropriate, unless I'm missing something.

jtkiesel: It seems that the `&&` version of the switch label "guarded" pattern is no more, and has been replaced with `when` in Java 21 (the first version that supports switch label pattern matching outside of a preview feature), which is scheduled to release tomorrow. So I will likely need to rethink this solution regardless. https://docs.oracle.com/javase/specs/jls/se21/html/jls-14.html#jls-SwitchLabel

jtkiesel:
#606 solves the issue described in #605 also, while simultaneously supporting the now-official "when" style of guards. Assuming that PR is merged, we can close this one.

PR #611, opened 2023-10-27
feat: align pattern matching with Java 21 spec

### LINKED ISSUE #605: Line break between instanceof and final

### LINKED ISSUE #610: Record pattern in exhaustive switch syntax unsupported
Record pattern appears in java 21 and prettier is unable to parse it

**Prettier-Java 2.3.1**

**Input:**

```java
package monad;

import java.util.Objects;
import java.util.function.Function;
import java.util.function.Predicate;

public sealed interface Maybe<T> {

  static <T> Maybe<T> of(T value) {
    if (value == null) {
      return new None<>();
    }
    return new Some<>(value);
  }

  default T valueOr(T defaultValue) {
    return switch (this) {
      case Some<T> some -> some.value();
      case None() -> defaultValue;
    };
  }

  default boolean isPresent() {
    return switch (this) {
      case Some<T> __ -> true;
      case None() -> false;
    };
  }

  default <R> Maybe<R> map(Function<T, R> mapper) {
    Objects.requireNonNull(mapper);

    return switch (this) {
      case Some(T value) -> of(mapper.apply(value)); // prettier fails here
      case None() -> new None<>();
    };
  }

  default Maybe<T> filter(Predicate<T> predicate) {
    Objects.requireNonNull(predicate);

    return switch (this) {
      case Some(T value) when predicate.test(value) -> new Some<>(value);
      default -> new None<>();
    };
  }
}

record Some<T>(T value) implements Maybe<T> {
}

record None<T>() implements Maybe<T> {
}
```

**Output:**

```sh
npx prettier --write src/main/java/monad/Maybe.java 
src/main/java/monad/Maybe.java
[error] src/main/java/monad/Maybe.java: Error: Sad sad panda, parsing errors detected in line: 34, column: 19!
[error] Expecting --> ')' <-- but found --> 'value' <--!
[error]         ->compilationUnit
[error]         ->ordinaryCompilationUnit
[error]         ->typeDeclaration
[error]         ->interfaceDeclaration
[error]         ->normalInterfaceDeclaration
[error]         ->interfaceBody
[error]         ->interfaceMemberDeclaration
[error]         ->interfaceMethodDeclaration
[error]         ->methodBody
[error]         ->block
[error]         ->blockStatements
[error]         ->blockStatement
[error]         ->statement
[error]         ->statementWithoutTrailingSubstatement
[error]         ->returnStatement
[error]         ->expression
[error]         ->ternaryExpression
[error]         ->binaryExpression
[error]         ->unaryExpression
[error]         ->primary
[error]         ->primaryPrefix
[error]         ->switchStatement
[error]         ->switchBlock
[error]         ->switchRule
[error]         ->switchLabel
[error]         ->caseOrDefaultLabel
[error]         ->caseLabelElement
[error]         ->caseConstant
[error]         ->ternaryExpression
[error]         ->binaryExpression
[error]         ->unaryExpression
[error]         ->primary
[error]         ->primarySuffix
[error]         ->methodInvocationSuffix
```

jtkiesel:
## What changed with this PR:

Aligns pattern matching parsing with the Java 21 spec, and attempts to improve the formatting of switch pattern matching. Record patterns are now supported, and the preview-style guards (`&&`) have been replaced with the official-style guards (`when`).

NOTE: This means that the `&&`-style guards will no longer be supported by the parser/printer. Because that style was only present as a preview feature (and has been removed as of Java 21 in favor of the official style), perhaps that is okay. I will defer to the project owners. If they decide we must continue supporting the pattern matching preview-style guards, I can attempt to modify this PR to support both simultaneously. My initial thought is that it would be incredibly easy to support both -- I should be able to simply allow a `Guard` to contain either a `When` or an `AndAnd`, instead of only a `When`.

### Input
```java
class Example {

  void example() {
    return switch (r) {
      case null, default -> 0;
      case MyRecord(A a) -> 0;
      case MyRecord(A a, B b) -> 0;
      case MyRecord(MyRecord(A a), B b) -> 0;
      case MyRecord(MyLongRecordTypeName(LongTypeName longVariableName, LongTypeName longVariableName), MyLongRecordTypeName(LongTypeName longVariableName, LongTypeName longVariableName)) -> 0;
      case MyRecord(LongTypeName longVariableName, LongTypeName longVariableName) -> 0;
      case MyRecord(LongTypeName longVariableName, LongTypeName longVariableName) when this.longVariableName > longVariableName && this.longVariableName > longVariableName -> 0;
      case MyRecord(LongTypeName longVariableName, LongTypeName longVariableName) when this.longVariableName > longVariableName && this.longVariableName > longVariableName -> longMethodName(longVariableName, longVariableName, longVariableName, longVariableName);
    };
  }
}
```

### Output
```java
class Example {

  void example() {
    return switch (r) {
      case null, default -> 0;
      case MyRecord(A a) -> 0;
      case MyRecord(A a, B b) -> 0;
      case MyRecord(MyRecord(A a), B b) -> 0;
      case MyRecord(
        MyLongRecordTypeName(
          LongTypeName longVariableName,
          LongTypeName longVariableName
        ),
        MyLongRecordTypeName(
          LongTypeName longVariableName,
          LongTypeName longVariableName
        )
      ) -> 0;
      case MyRecord(
        LongTypeName longVariableName,
        LongTypeName longVariableName
      ) -> 0;
      case MyRecord(
        LongTypeName longVariableName,
        LongTypeName longVariableName
      ) when (
        this.longVariableName > longVariableName &&
        this.longVariableName > longVariableName
      ) -> 0;
      case MyRecord(
        LongTypeName longVariableName,
        LongTypeName longVariableName
      ) when (
        this.longVariableName > longVariableName &&
        this.longVariableName > longVariableName
      ) -> longMethodName(
        longVariableName,
        longVariableName,
        longVariableName,
        longVariableName
      );
    };
  }
}
```

Simultaneously fixes an `instanceof` pattern matching printer bug.

### Input
```java
class Example {

  public boolean test(final Object obj) {
    return (
      obj instanceof
      final Integer x &&
      (x == 5 || x == 6 || x == 7 || x == 8 || x == 9 || x == 10 || x == 11)
    );
  }
}
```

### Output
```java
class Example {

  public boolean test(final Object obj) {
    return (
      obj instanceof final Integer x &&
      (x == 5 || x == 6 || x == 7 || x == 8 || x == 9 || x == 10 || x == 11)
    );
  }
}
```

## Relative issues or prs:

Closes #605
Closes #610
Obsoletes #606


jtkiesel:
Fixed a printer indentation bug with (and added a test for) nested record patterns.

packages/java-parser/src/productions/classes.js
```diff
@@ -202,9 +202,12 @@ function defineRules($, t) {
   // https://docs.oracle.com/javase/specs/jls/se16/html/jls-8.html#jls-VariableDeclaratorList
   $.RULE("variableDeclaratorList", () => {
     $.SUBRULE($.variableDeclarator);
-    $.MANY(() => {
-      $.CONSUME(t.Comma);
-      $.SUBRULE2($.variableDeclarator);
+    $.MANY({
+      GATE: () => !tokenMatcher(this.LA(3).tokenType, t.Identifier),
```
clementdessoude, line 206: It took me some time to understand why it could cause a problem. Could we add add a comment here in order to make it clear why we need this (to distinguish from patternList) ? And maybe a test sample in the parser's tests ?

For instance:
```ts
it ("should parse pattern list", () => {
  const input = `A a, B b`;
  expect(() => javaParser.parse(input, "patternList")).to.not.throw();
});
```

packages/java-parser/src/productions/expressions.js
```diff
@@ -561,33 +561,49 @@ function defineRules($, t) {
...
-
-  $.RULE("primaryPattern", () => {
     $.OR([
       {
-        ALT: () => {
-          $.CONSUME(t.LBrace);
-          $.SUBRULE($.pattern);
-          $.CONSUME(t.RBrace);
-        }
+        GATE: () => this.BACKTRACK_LOOKAHEAD($.recordPattern),
```
clementdessoude, line 568: I think we could switch the two options: as there is recursion in the pattern/recordPattern definition, I would suggest backtracking on the typePattern instead

packages/java-parser/src/productions/classes.js
jtkiesel: Done!

packages/java-parser/src/productions/expressions.js
jtkiesel: Good point! Done.

packages/prettier-plugin-java/test/unit-test/pattern-matching/_output.java
```diff
@@ -19,43 +19,92 @@ static String formatter(Object o) {
...
+      case Buyer b when this.bestPrice > b.bestPrice -> true;
+      case Buyer b when
         this.bestPrice > b.bestPrice &&
         this.bestPrice > b.bestPrice &&
         this.bestPrice > b.bestPrice &&
-        this.bestPrice > b.bestPrice -> true;
-      case Buyer b &&
+        this.bestPrice > b.bestPrice
+      -> true;
+      case Buyer b when
```
clementdessoude, line 43: I'm wondering if it would be better if we add parenthesis on break ?

It would be something like that:

```java
void test(Buyer other) {
  return switch (other) {
    case null -> true;
    case Buyer b when this.bestPrice > b.bestPrice -> true;
    case Buyer b when this.bestPrice > b.bestPrice -> {
      return true;
    }
    case Buyer titi when (
      this.bestPriceaaaaaaaazzzzzaaaaaaaaaq > b.bestPrice
    ) -> true;
    case Buyer b when (
      this.bestPrice > b.bestPrice &&
      this.bestPrice > b.bestPrice &&
      this.bestPrice > b.bestPrice &&
      this.bestPrice > b.bestPrice
    ) -> true;
    case Buyer b when (
      this.bestPrice > b.bestPrice &&
      this.bestPrice > b.bestPrice &&
      this.bestPrice > b.bestPrice &&
      this.bestPrice > b.bestPrice
    ) -> {
      return true;
    }
    case Buyer toto when (
      this.bestPrice > b.bestPrice &&
      this.bestPrice > b.bestPrice &&
      this.bestPrice > b.bestPrice &&
      this.bestPrice > b.bestPrice
    ) -> true;
    case Buyer b when (
      this.bestPrice > b.bestPrice &&
      this.bestPrice > b.bestPrice &&
      this.bestPrice > b.bestPrice &&
      this.bestPrice > b.bestPrice
    ) -> {
      return true;
    }
    default -> false;
  };
}
```

I was trying to do it, but I still have one issue if the line is breaking just after the arrow :

```java
void test(Buyer other) {
  return switch (other) {
    case Buyer titi when (
      this.bestPriceaaaaaaaazzzzzaaaaaaaaaq > b.bestPrice
    ) -> true;
    case Buyer titi when this.bestPriceaaaaaazzzaaaaaaaaaq > b.bestPrice ->
    true;
  };
}
```

packages/prettier-plugin-java/src/printers/blocks-and-statements.ts
```diff
@@ -331,12 +335,25 @@ export class BlocksAndStatementPrettierVisitor extends BaseCstPrettierPrinter {
...
-    return join(" ", [switchLabel, ctx.Arrow[0], caseInstruction]);
+    return concat([
+      group([switchLabel, separator, group([ctx.Arrow[0], line])]),
+      caseInstruction
+    ]);
   }
 
   caseConstant(ctx: CaseConstantCtx) {
-    return this.visitSingle(ctx);
+    const constant = this.visitSingle(ctx);
```
clementdessoude, line 351: I'm not sure why you need to do this ? Do you have a test sample to reproduce ?

jtkiesel: I think this was accidentally leftover from #606. However, something similar to this change seems necessary to properly indent `CaseConstant`s with parentheses (for example, a `BinaryExpression` wrapped with parentheses can be used as a `CaseConstant`). As I'm working on your suggestion in https://github.com/jhipster/prettier-java/pull/611#discussion_r1375387354, I'm trying to make a similar change for `CaseConstant`s (i.e. adding parentheses if they wrap).

jtkiesel: Never mind, this ended up being a lot more complicated than I thought. Since it's unrelated to the purpose of the PR (Java 21 pattern matching/guards support), I'm going to get rid of these changes and leave this improvement for another PR.

packages/prettier-plugin-java/test/unit-test/pattern-matching/_output.java
jtkiesel: I like this a lot better! It also helped simplify the printer logic in `switchRule`. Just pushed this improvement, let me know what you think.

packages/prettier-plugin-java/test/unit-test/pattern-matching/_input.java
```diff
@@ -13,23 +13,40 @@ static String formatter(Object o) {
...
-            case Buyer b && this.bestPrice > b.bestPrice -> true;
-            case Buyer b && this.bestPrice > b.bestPrice -> {
+            case Buyer b when this.bestPrice > b.bestPrice -> true;
+            case Buyer b when this.bestPrice > b.bestPrice -> {
                 return true;
             }
-            case (Buyer b && this.bestPrice > b.bestPrice) -> true;
-            case Buyer b && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice -> true;
-            case Buyer b && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice -> {
+            case Buyer b when this.bestPrice > b.bestPrice -> true;
```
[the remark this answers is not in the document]
jtkiesel: Thanks! Removed the duplicate.

packages/prettier-plugin-java/test/unit-test/pattern-matching/_input.java
```diff
@@ -13,23 +13,40 @@ static String formatter(Object o) {
...
+            case Buyer b when this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice -> true;
+            case Buyer b when this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice -> {
                 return true;
             }
-            case (Buyer b && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice) -> true;
-            case (Buyer b && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice) -> {
+            case Buyer b when this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice -> true;
+            case Buyer b when this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice && this.bestPrice > b.bestPrice -> {
```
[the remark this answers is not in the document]
jtkiesel: Thanks! Removed the duplicates.

clementdessoude:
Thanks @jtkiesel !

PR #613, opened 2023-10-29
chore: upgrade pipeline tests to JHipster 8, Java 17, and Node.js 18/20

### LINKED ISSUE #609: Migrate e2e pipeline tests to use Java 17
For the newest Jhipster 8 version, Java default version is now Java 17.

Following https://github.com/jhipster/prettier-java/pull/588, E2E pipeline testing needs to be update to use Java 17 and may be use Node 18 as minor node version (Node 14 and 16 are deprecated).

jtkiesel:
## What changed with this PR:

Upgrades the pipeline tests to JHipster 8, Java 17, and Node.js 18/20. Also upgraded GitHub Actions and repos used for testing to their latest versions (some were necessary upgrades, such as Spring Boot/Framework, others were not).

## Relative issues or prs:

Closes #609

### DIFF
--- a/packages/prettier-plugin-java/scripts/clone-samples.js
+++ b/packages/prettier-plugin-java/scripts/clone-samples.js
@@ -9,65 +9,65 @@ const samplesDir = path.resolve(__dirname, "../samples");
 const core = [
   {
     repoUrl: "https://github.com/jhipster/jhipster-sample-app",
-    branch: "v7.9.3"
+    branch: "v8.0.0-rc.1"
   },
   {
-    repoUrl: "https://github.com/jhipster/jhipster",
-    branch: "main"
+    repoUrl: "https://github.com/jhipster/jhipster-bom",
+    branch: "8.0.0-rc.1"
   },
   {
     repoUrl: "https://github.com/spring-projects/spring-boot.git",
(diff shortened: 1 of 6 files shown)


packages/prettier-plugin-java/scripts/clone-samples.js
```diff
@@ -9,65 +9,65 @@ const samplesDir = path.resolve(__dirname, "../samples");
 const core = [
   {
     repoUrl: "https://github.com/jhipster/jhipster-sample-app",
-    branch: "v7.9.3"
+    branch: "main"
```
clementdessoude, line 12: I prefer to keep a fixed version, in order to have more reproducible builds/tests

clementdessoude (changes requested):
Hello @jtkiesel, I just have one remarks about using fixed version for jhipster samples, but otherwise, it's ok for me !

packages/prettier-plugin-java/scripts/clone-samples.js
jtkiesel: Agreed! I originally decided to go with the way things were prior to #588, in case there was a reason that was originally done that I wasn't aware of. Just pushed this change! 👍

clementdessoude (approved):

Issue #495, opened 2021-08-14
How to convert Spaces to tabs


hdpklm:
hello all

when i format java file all tabs are conterted to spaces.
how to fix the problem ???

thank you all

gmarziou:
Edit your` .prettierrc` file and set `useTabs: true`

PR #622, opened 2023-11-17
Add information on how to change indent

koppor:
## What changed with this PR:

More beginner-friendly documentation.

When coming from google-java-format --ASOP or palantir-java-format, it is not that obvious how to use prettier-java. Adding some lines of documentation makes it obvious (and thus this project is more beginner-friendly)

## Relative issues or prs:

Fixes https://github.com/jhipster/prettier-java/issues/495.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,13 +141,7 @@ For using four spaces:
 tabWidth: 4
 ```
 
-For using tabs:
-
-```yaml
-useTabs: true
-```
-
-For more details see <https://prettier.io/docs/en/configuration.html>.
+For more configuration options such as using tabs, maximum line length, and more see <https://prettier.io/docs/en/configuration.html>.
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -131,6 +131,24 @@ You can directly import the config into IntelliJ Idea for instance:
 
 ![Import Checkstyle configuration](./docs/checkstyle/import-checkstyle-configuration.gif)
 
+## Indent configuration
+
+In `.prettierrc.yaml`, you can configure the indent:
+
+For using four spaces:
+
+```yaml
+tabWidth: 4
+```
+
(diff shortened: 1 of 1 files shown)


DanielFran (approved):

murdos:
I think it would be better to redirect to the prettier documentation rather than documenting again each configuration parameter.

koppor:
> I think it would be better to redirect to the prettier documentation rather than documenting again each configuration parameter.

My two main reasons:

I aimed for documenting the WTF configuration parameters, not all. Sun Java style is 4 spaces. Nearly all Java code has 4 saves. Default configuration is 2 spaces, which might cause WTF moments.

The prettier configuration is not that obvious for impatient people outside the NodeJS world.

I am open to remove the tab configuration - this is only needed for projects wanting to be more accessible. This might be not worth to make it prominent. Persons caring for exceptional groups are surely able to read and interpet linked documentation.

Some background material,:

Tabs are objectively better than spaces (due to accessibility reasons): https://gomakethings.com/tabs-are-objectively-better-than-spaces/
   >  Another accessibility reason for tabs: when using a braille display, each space takes up one character cell, so indenting with four spaces eats up four cells. Indenting three times with four spaces each eats up 12 characters already. Tabs only take one character cell each, so three indents = three character cells used.
   >
   > Given many, if not most, braille displays are 40 cells long, and that 70-to 80-cell displays are considered a luxury, and you can see how cutting 12 space characters down to 3 tab characters could be really, really helpful.
- Another article arguing for accessibility: https://alexandersandberg.com/articles/default-to-tabs-instead-of-spaces-for-an-accessible-first-environment/

koppor:
I got the idea of having the indent configuration in the main README.md from `spring-javaformat` at https://github.com/spring-io/spring-javaformat#about-the-conventions.

PR #625, opened 2023-11-21
java-parser: New entrypoint `lexAndParse` to return both tokens and CST.

max-schaefer:
## What changed with this PR:

Provide an entrypoint that exposes both the CST and the underlying token array.

## Alternatives

The token array could also be accessed through the `parser` object, which we could export instead, but that seems like an internal implementation detail which perhaps shouldn't be part of the API.

Finally, we could consider exporting the lexer, but then if I want to get both the CST and the tokens I have to lex twice, which seems wasteful.

### DIFF
--- a/packages/java-parser/api.d.ts
+++ b/packages/java-parser/api.d.ts
@@ -30,6 +30,14 @@ export declare type CstChildrenDictionary = {
   [identifier: string]: CstElement[];
 };
 
+export function lexAndParse(
+  text: string,
+  startProduction?: string
+): {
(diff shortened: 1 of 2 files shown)


DanielFran:
@max-schaefer can you sign the CLA?

max-schaefer:
Yep, I'm looking into it (currently checking with my employer whether they are OK with me signing it).

max-schaefer:
CLA is signed.

clementdessoude:
Thanks for this addition @max-schaefer !

Issue #462, opened 2021-04-14
Cannot import into Nuxt project


cameronweibel:
# Description
Importing this plugin into a Nuxt project and building with `npm run start (nuxt generate and nuxt start)` is not possible. 

# Reproduction 
Creating a page in Nuxt with the following code:
```
<template>
  <h1>Hi</h1>
</template>

<script>
import parserJava from "prettier-plugin-java";
</script>
```

And building, navigating to page, results in an error of:
`TypeError: Cannot read property '1' of null at vn (a7816b4.js:2) at e.exports.e.initRecognizerEngine (a7816b4.js:2) at new e (a7816b4.js:2) at new e.exports (a7816b4.js:2) at Object.<anonymous> (a7816b4.js:2) at l (094da21.js:1) at Object.<anonymous> (a7816b4.js:2) at l (094da21.js:1) at Object.<anonymous> (a7816b4.js:2) at l (094da21.js:1)`

[nuxt-static-error.zip](https://github.com/jhipster/prettier-java/files/6313384/nuxt-static-error.zip)
Run `npm i` followed by `npm run start` in this project to see the issue. If you run `npm run dev` it works.

cameronweibel:
Apparently - "This is not a Nuxt-specific issue. The error is being thrown by chevrotain, a dependency of prettier-plugin-java. It is inspecting the source code of a class with a RegExp to try to find a function name - which, because it's built, is different from development mode."

cameronweibel:
For anybody else reading this with the same issue, I fixed it with the following build options in nuxt.config.js:

  build: {
      terser: {
          parallel: true,
          cache: false,
          sourceMap: false,
          extractComments: {
              filename: "LICENSES",
          },
          terserOptions: {
              output: {
                  comments: /^\**!|@preserve|@license|@cc_on/,
              },
              keep_fnames: true, <- This lines fixes it
          },
      },
  },

It would be ideal to not have to use this workaround though.

Shaolans:
I will be adding this in the documentation.
Thank you for giving this workaround

PR #628, opened 2023-12-30
docs: create Docusaurus website

### LINKED ISSUE #193: Create Demos/Playground

jtkiesel:
## What changed with this PR:

Prettier Java now has a Docusaurus website, including an interactive playground that allows users to see how it works. I am currently self-hosting it [here](https://jtkiesel.github.io/prettier-java/). ~I have made this a draft PR because the playground currently depends on #616 being merged first, or else #462 prevents it from working properly (that issue seems to have been fixed by newer Chevrotain versions).~

## Relative issues or prs:

Closes #193


website/src/pages/playground/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+                />{" "}
+                --use-tabs
+              </label>
+            </details>
+            <details open>
+              <summary>Java</summary>
+              <label title="Print trailing commas wherever possible when multi-line.">
+                --trailing-comma{" "}
+                <select
+                  disabled={true} // this option isn't seen by prettier-plugin-java when passed programmatically
```
clementdessoude, line 95: It's strange that the other options are working but not this one, I wonder what causes this issue 🤔

clementdessoude (commented):
Waouh, looks great ! The playground worked fine on Google Chrome but does not seems to work on Firefox, do you know why ?

jtkiesel:
> The playground worked fine on Google Chrome but does not seems to work on Firefox, do you know why ?

@clementdessoude Hmm. Which version of Firefox? I tried it on the latest version for macOS (121.0.1) and it seemed to work properly. For what it's worth, I also tried it on Safari and it seemed to work well.

website/src/pages/playground/index.tsx
jtkiesel: Figured this out -- just had to add `trailingComma` as an option in [`options.js`](https://github.com/jhipster/prettier-java/pull/628/files#diff-e239c88e07ba045568350e8839476ffe09f9f6d2c986a60e83b04aa188ea3203).

jtkiesel:
Pushed an update that implements a "state hash" feature on the playground (very similar to the one on Prettier's playground), which enables the sharing of links to the playground with the code and all options persisted.

website/src/pages/playground/index.tsx
clementdessoude: Great !

clementdessoude:
> > The playground worked fine on Google Chrome but does not seems to work on Firefox, do you know why ?
> 
> @clementdessoude Hmm. Which version of Firefox? I tried it on the latest version for macOS (121.0.1) and it seemed to work properly. For what it's worth, I also tried it on Safari and it seemed to work well.

I'm on Firefox 121.0.1 (it seems that it is the latest version)

Screenshot 2024-01-22 at 09 50 43

As you see in the screenshot, I don't see any code in the formatted section. I don't understand why for the moment, but it seems that the width property is not taken into account for the second editor :/

image

jtkiesel:
@clementdessoude Thanks for the screenshots! I realize now that when I opened the website in Firefox, I had it open in a small window, so the editors were rendered on top of each other, rather than next to each other, which happened to work. Once I expanded my browser window, I also saw the issue.

I just pushed a fix. For some reason, on Firefox `flex-flow: wrap;` wasn't working the same as it was on Chrome. On Firefox, it didn't seem to allow the 2nd flex item (the 2nd editor) to fully reach the right edge of the viewport. If I changed the width of the editors to be `0.1px` smaller, it worked, but then there was an odd tiny gab on the right side. I ended up changing it to remove `flex-flow: wrap;` and simply change the `flex-direction` based on the screen width, and it now seems to work properly on both Chrome and Firefox.

jtkiesel:
@clementdessoude Thank you for merging this! When you have a moment, could you follow the [deployment steps](https://github.com/jhipster/prettier-java/tree/main/website#deployment) on a direct clone (not a fork) of this repo? Doing that will create a `gh-pages` branch on this repo containing the compiled website, and GitHub will use it to host the website at https://jhipster.github.io/prettier-java/.

clementdessoude:
It's done :)

jtkiesel:
@clementdessoude Thanks! Looks like it worked. 🎉 As a heads up, in order for the website to use the latest version of the plugin (in the playground), those deployment steps should be run (following a local build of the plugin) after every release of the plugin. I'm not sure if it's possible to automate that process? If you use the [deploy.sh](https://github.com/jhipster/prettier-java/blob/main/scripts/deploy.sh) script, perhaps we can/should add it to that?

PR #632, opened 2024-01-09
feat: avoid breaking on certain method chains and arguments

### LINKED ISSUE #626: Don't split a short line static method to two lines.
putting `Try` and `of` on separate lines here seems weird.

**Input:**

```java
    private static <T extends Group> Function<Constructor<?>, T> createInstance(Group entity) {
        return ctor -> Try.of(() -> {
                    @SuppressWarnings("unchecked")
                    var ng = (T) ctor.newInstance(entity.getId(), entity.getSystemGenerated(), entity.getVersionKey());
                    return ng;
                })
                .getOrElseThrow(ex -> new RuntimeException(ex));
    }
```

**Output:**

```java
  private static <T extends Group> Function<Constructor<?>, T> createInstance(Group entity) {
    return ctor ->
      Try
        .of(() -> {
          @SuppressWarnings("unchecked")
          var ng = (T) ctor.newInstance(entity.getId(), entity.getSystemGenerated(), entity.getVersionKey());
          return ng;
        })
        .getOrElseThrow(ex -> new RuntimeException(ex));
  }
```

**Expected behavior:**

```java
  private static <T extends Group> Function<Constructor<?>, T> createInstance(Group entity) {
    return ctor ->
      Try.of(() -> {
          @SuppressWarnings("unchecked")
          var ng = (T) ctor.newInstance(entity.getId(), entity.getSystemGenerated(), entity.getVersionKey());
          return ng;
        })
        .getOrElseThrow(ex -> new RuntimeException(ex));
  }
```

or perhaps even this since the return is implicit.

```java
    return ctor -> Try.of(() -> {
```

**Configuration:**

```json
  "dependencies": {
    "prettier": "^3.1.1",
    "prettier-plugin-java": "^2.5.0"
  }
```

```yaml
printWidth: 120
plugins:
  - prettier-plugin-java
```

```ini
root = true

[*]
trim_trailing_whitespace = true
end_of_line = lf
insert_final_newline = true
charset = utf-8
indent_style = space
indent_size = 2

[*.bat]
end_of_line = crlf
```

jtkiesel:
## What changed with this PR:

Method chains and arguments are now broken more similarly to Prettier JavaScript. Method invocations assumed to be static (i.e. those on capitalized identifiers) are now never broken, short method chains prefer to be broken on arguments instead, and argument lists containing a single lambda with body at the end prefer not to break.

### Input
```java
class Example {

  Class<?> aaaaaaaaaaaaaaaa = new Aaaaaaaaaaaaaaaa<Bbbbbbbbbbbbbbbb>[1]
    .getClass();
  Class<?> aaaaaaaaaaaaaaaa = new Aaaaaaaaaaaaaaaa[] { new Aaaaaaaaaaaaaaaa() }
    .getClass();

  void example() {
    List
      .of(
        new double[][] { 1, 2, 3, 4.1, 5.6846465 },
        new double[][] { 1, 2, 3, 4.1, 5.6846465 },
        new double[][] { 1, 2, 3, 4.1, 5.6846465 }
      )
      .toArray(double[][]::new);

    a.of(
      b,
      c,
      d,
      e -> {
        return f;
      }
    );

    Stream
      .of(1, 2)
      .map(n -> {
        // testing method
        return n * 2;
      })
      .collect(Collectors.toList());

    return Object
      .something()
      .more()
      .and()
      .that()
      .as()
      .well()
      .but()
      .not()
      .something();
  }
}
```

### Output
```java
class Example {

  Class<?> aaaaaaaaaaaaaaaa = new Aaaaaaaaaaaaaaaa<
    Bbbbbbbbbbbbbbbb
  >[1].getClass();
  Class<?> aaaaaaaaaaaaaaaa = new Aaaaaaaaaaaaaaaa[] {
    new Aaaaaaaaaaaaaaaa(),
  }.getClass();

  void example() {
    List.of(
      new double[][] { 1, 2, 3, 4.1, 5.6846465 },
      new double[][] { 1, 2, 3, 4.1, 5.6846465 },
      new double[][] { 1, 2, 3, 4.1, 5.6846465 }
    ).toArray(double[][]::new);

    a.of(b, c, d, e -> {
      return f;
    });

    Stream.of(1, 2)
      .map(n -> {
        // testing method
        return n * 2;
      })
      .collect(Collectors.toList());

    return Object.something()
      .more()
      .and()
      .that()
      .as()
      .well()
      .but()
      .not()
      .something();
  }
}
```

## Relative issues or prs:

Closes #626


packages/prettier-plugin-java/src/printers/expressions.ts
```diff
@@ -347,37 +384,17 @@ export class ExpressionsPrettierVisitor extends BaseCstPrettierPrinter {
...
-            ) &&
-            // indent when lambdaExpression
-            ctx.primarySuffix[0].children.methodInvocationSuffix &&
-            ctx.primarySuffix[0].children.methodInvocationSuffix[0].children
-              .argumentList &&
-            ctx.primarySuffix[0].children.methodInvocationSuffix[0].children
-              .argumentList[0].children.expression &&
-            ctx.primarySuffix[0].children.methodInvocationSuffix[0].children
-              .argumentList[0].children.expression[0].children
-              .lambdaExpression === undefined
```
clementdessoude, line 375: It's great we could get rid of this part !

packages/prettier-plugin-java/src/printers/expressions.ts
```diff
@@ -324,10 +321,50 @@ export class ExpressionsPrettierVisitor extends BaseCstPrettierPrinter {
...
+      countMethodInvocation <= 1 &&
+      [
+        arrayCreationExpression?.classOrInterfaceType?.[0].children.classType[0]
+          .children.typeArguments,
+        arrayCreationExpression?.arrayCreationExplicitInitSuffix?.[0].children
+          .arrayInitializer[0].children.variableInitializerList,
+        classInstanceCreationExpression?.classOrInterfaceTypeToInstantiate[0]
+          .children.typeArgumentsOrDiamond?.[0].children.typeArguments,
+        classInstanceCreationExpression?.argumentList
+      ].some(breakablePart => breakablePart);
```
clementdessoude, line 340: We maybe could extract a isBreakableNewExpression function to lighten the cognitive load on this function ?

packages/prettier-plugin-java/src/printers/expressions.ts
```diff
@@ -324,10 +321,50 @@ export class ExpressionsPrettierVisitor extends BaseCstPrettierPrinter {
...
+      ...(fqnOrRefType?.fqnOrRefTypePartRest ?? [])
+    ];
+    const nextToLastIdentifier =
+      fqnOrRefTypeParts[fqnOrRefTypeParts.length - 2]?.children
+        .fqnOrRefTypePartCommon[0].children.Identifier?.[0].image;
+    const firstMethodInvocation = ctx.primarySuffix
+      ?.map(suffix => suffix.children.methodInvocationSuffix?.[0].children)
+      .find(methodInvocationSuffix => methodInvocationSuffix);
+    const isCapitalizedIdentifier =
+      nextToLastIdentifier && /^\p{Lu}/u.test(nextToLastIdentifier);
```
clementdessoude, line 354: Could we use Uppercase_Letter instead of Lu, I found it more explicit ?

packages/prettier-plugin-java/src/printers/expressions.ts
```diff
@@ -324,10 +321,50 @@ export class ExpressionsPrettierVisitor extends BaseCstPrettierPrinter {
...
+      ].some(breakablePart => breakablePart);
+    const fqnOrRefType =
+      ctx.primaryPrefix[0].children.fqnOrRefType?.[0].children;
+    const fqnOrRefTypeParts = [
+      fqnOrRefType?.fqnOrRefTypePartFirst[0],
+      ...(fqnOrRefType?.fqnOrRefTypePartRest ?? [])
+    ];
+    const nextToLastIdentifier =
+      fqnOrRefTypeParts[fqnOrRefTypeParts.length - 2]?.children
+        .fqnOrRefTypePartCommon[0].children.Identifier?.[0].image;
```
clementdessoude, line 349: WDYT of extracting a nextToLastIdentifier function ?

packages/prettier-plugin-java/src/printers/expressions.ts
clementdessoude: I would also be in favor of extracting a method here, even if it is a small one

packages/prettier-plugin-java/test/unit-test/lambda/_output.java
```diff
@@ -143,6 +140,49 @@ public void testConstructor() {
...
+        );
+        return ng;
+      }).getOrElseThrow(ex -> new RuntimeException(ex));
+  }
+
+  void singleLambdaWithBlockLastArgument() {
+    a.of(b, c, d, e -> {
+      return f;
+    });
+  }
```
clementdessoude, line 163: I think there is an issue with formatting this snippet

**Input:**
```java
class T {
    public void method() {
        a.of(
            aaaaaaaaaaaaaaaaaaaaaaaaaa,
            bbbbbbbbbbbbbbbbbbbbbbbbbb,
            cccccccccccccccccccccccccc,
            e -> {
                return f;
            }
        );
    }
}

```
**Output:**
```java
class T {

  public void method() {
    a.of(aaaaaaaaaaaaaaaaaaaaaaaaaa,
    bbbbbbbbbbbbbbbbbbbbbbbbbb,
    cccccccccccccccccccccccccc,
    e -> {
      return f;
    });
  }
}

```

packages/prettier-plugin-java/src/printers/expressions.ts
```diff
@@ -324,10 +321,50 @@ export class ExpressionsPrettierVisitor extends BaseCstPrettierPrinter {
...
+    const firstMethodInvocation = ctx.primarySuffix
+      ?.map(suffix => suffix.children.methodInvocationSuffix?.[0].children)
+      .find(methodInvocationSuffix => methodInvocationSuffix);
+    const isCapitalizedIdentifier =
+      nextToLastIdentifier && /^\p{Lu}/u.test(nextToLastIdentifier);
+    const shouldBreakBeforeFirstMethodInvocation =
```
clementdessoude, line 355: In the end, I ended up with something like that, WDYT ?

```ts
    ...
    const { newExpression, nextToLastIdentifier, firstMethodInvocation } = extract(ctx);
    const isCapitalizedIdentifier = isCapitalized(nextToLastIdentifier);

    const shouldBreakBeforeFirstMethodInvocation =
      countMethodInvocation > 1 &&
      !(isCapitalizedIdentifier) &&
      firstMethodInvocation !== undefined;
    ...
```

PS: the name of the extract method should be changed :)

clementdessoude (changes requested):
Great work @jtkiesel ! 

There is one issue with formatting method invocation with lambda, but otherwise, seems to be working fine !

If possible, it would be great to extract some functions at the beginning of the expression method, as this method brings a lot of cognitive load

packages/prettier-plugin-java/test/unit-test/lambda/_output.java
jtkiesel: Thanks for catching this! I was able to fix this one without much issue. However, it lead me to find another that has me stumped at the moment:
```java
Class T {

  void singleLambdaWithBlockLastArgumentAndLongLambdaArguments() {
    a.of(b,
      (
        cccccccccccccccccccccccccc,
        dddddddddddddddddddddddddd,
        eeeeeeeeeeeeeeeeeeeeeeeeee
      ) -> {
        return f;
      }
    );
  }
}
```
I can't use `ifBreak` to conditionally break before `b,` if the lambda's argument list breaks, because `ifBreak` requires the provided `groupId` to be one that has already been printed (i.e. it must be a `groupId` that appears before the `ifBreak`). The only solution I've thought of so far is grouping the lambda's argument list with the "parent" argument list (in this case, `b`), so that the lambda argument list breaking also breaks the "parent" argument list. However, that seems rather ugly and I'm not even sure if something like that is possible.

packages/prettier-plugin-java/src/printers/expressions.ts
jtkiesel: Since I extracted `isCapitalizedIdentifier` out into its own function and `nextToLastIdentifier` was only used there, I simply moved it into that function. Let me know if you'd still like that broken out further.

packages/prettier-plugin-java/test/unit-test/lambda/_output.java
jtkiesel: I was able to resolve both of these. The latter required some additional research, but looking into Prettier's own source code, I realized they were solving this for JS/TS by using `conditionalGroup`, and I was able to do the same.

packages/prettier-plugin-java/src/printers/expressions.ts
jtkiesel: I ended up solving this via individual functions. Let me know what you think.

jtkiesel:
@clementdessoude I was able to solve the issues & refactored to reduce the cognitive complexity a bit. I was also able to simplify a few things in the process. Let me know if you think the `primary` function could use any additional refactoring.

PR #693, opened 2024-10-24
fix: move prettier from `dependencies` to `devDependencies`

ntnyq:
## What changed with this PR:

This PR moves prettier from `dependencies` to `devDependencies` and adds it to `peerDependencies`.


jtkiesel:
Thanks for this correction! 🎉

PR #714, opened 2025-01-17
fix: restore require support for VS Code compatibility

### LINKED ISSUE #711: prettier-plugin-java not working in prettier-vscode anymore
Since the last version update (2.6.6) prettier-plugin-java can not be loaded by prettier-vscode (https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) anymore.

The culprit seems to be this change: https://github.com/jhipster/prettier-java/commit/2e7a828a05d1a396cec697d7dce2d9cb92e3d5ab
It changed the exports field in package.json and only includes "types" and "import". But the Vscode plugin tries to load the modules using require.resolve('prettier-plugin-java').
Node throws an exception _No "exports" main defined in .../node_modules/prettier-plugin-java/package.json_

Changing the exports to include a require element, fixes the problem.

```
"exports": {
    ".": {
      "types": "./index.d.ts",
      "import": "./dist/index.js",
      "require": "./dist/index.js"
    }
  }
```

jtkiesel:
## What changed with this PR:

Support for require is restored for compatibility with the Prettier VS Code plugin.

## Relative issues or prs:

Closes #711

### DIFF
--- a/packages/prettier-plugin-java/package.json
+++ b/packages/prettier-plugin-java/package.json
@@ -4,10 +4,8 @@
   "description": "Prettier Java Plugin",
   "type": "module",
   "exports": {
-    ".": {
-      "types": "./index.d.ts",
-      "import": "./dist/index.js"
-    }
+    "types": "./index.d.ts",
+    "default": "./dist/index.js"
   },
   "files": [
     "dist",


packages/prettier-plugin-java/package.json
```diff
@@ -6,7 +6,7 @@
   "exports": {
     ".": {
       "types": "./index.d.ts",
-      "import": "./dist/index.js"
+      "default": "./dist/index.js"
     }
   },
```
clementdessoude, line 11: I think we can remove the "." part and put directly the "types" and "default" keys at the root of "exports"

```json
"exports": {
    "types": "./index.d.ts",
    "default": "./dist/index.js"
},
```

jtkiesel: Good point! I confirmed that still fixes the issue, and updated my PR.

PR #715, opened 2025-01-20
Fix/712 handle comments with static invocations

### LINKED ISSUE #712: Formatting not stable with comment between class and static function call
prettier: 3.4.2
pretier-plugin-java: 2.6.6

```bash
$ cat initial.java
package main;

class Main {

    return MyClass
        // comment
        .doSomething();
  }
}
package main;

class Main {

        return MyClass// comment
        .doSomething();
    }
}
package main;

class Main {

        return MyClass.doSomething(); // comment
    }
}
```

Maybe related #592

clementdessoude:
## What changed with this PR:

Fix unstable formattings with comments inside fqn parts. 
#713 was an attempt to fix the #712 issue but I realized afterwards that some other edge cases were remaining

## Example

See modified unit test files to see some examples

## Relative issues or prs:

Fix #712

### DIFF
--- a/packages/prettier-plugin-java/src/printers/expressions.ts
+++ b/packages/prettier-plugin-java/src/printers/expressions.ts
@@ -467,8 +467,9 @@ export class ExpressionsPrettierVisitor extends BaseCstPrettierPrinter {
           : concat([softline, dots[0]]);
         return rejectAndConcat([
           indent(
-            rejectAndJoin(separator, [
+            rejectAndConcat([
               fqnOrRefTypePartFirst,
+              separator,
               rejectAndJoinSeps(dots.slice(1), fqnOrRefTypePartRest),
               dims
             ])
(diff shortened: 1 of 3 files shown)


packages/prettier-plugin-java/test/unit-test/bug-fixes/_input.java
```diff
@@ -7,6 +7,10 @@ void process(
         Map.@NonNull Entry<String, SkeletonConfiguration> entry,
         @NonNull Map<String, Object> context
     ) {}
+
+    void process() {
+        coco.tata[]()();
```
jtkiesel, line 12: This isn't valid Java syntax, right? Does something like this still test the edge case you were hoping to account for?
```suggestion
        coco.tata[0]();
```

clementdessoude: You're right, I was too much on what was allowed by java-parser ! I removed this test case as it won't happen in real life.

Issue #660, opened 2024-04-05
formatWithCursor and --cursor-offset seems not work


nonsense-j:
**Prettier-Java 2.6.0**
### Detail
I want to get the index change of given cursor when formatting Java code. I noticed that it can be done with `formatWithCursor`(API) or `--cursor-offset`(CLI). However, whatever valid offset I try, the result will always be 0. Besides, I have also tried with formatting babel code (builtin in Prettier), which behaved correctly. I wonder how I can fix this. Is this a feature unfinished?

I will show you the example with API (CLI also fails extractly the same)
### API Usage with `formatWithCursor`
```js
import * as prettier from "prettier";
import * as prettierPluginJava from "prettier-plugin-java";

const javaText = `
  public class HelloWorldExample{public static void main(String args[]){
    System.out.println("Hello World !");
  }
}
`;

const formattedText = await prettier.formatWithCursor(javaText, {
  cursorOffset: 5,
  parser: "java",
  plugins: [prettierPluginJava.default],
});
```

**Output:**
```java
{
  formatted: 'public class HelloWorldExample {\n' +
    '\n' +
    '  public static void main(String args[]) {\n' +
    '    System.out.println("Hello World !");\n' +
    '  }\n' +
    '}\n',
  cursorOffset: 0,
  comments: []
}
```

**Expected output:**
```java
{
  formatted: 'public class HelloWorldExample {\n' +
    '\n' +
    '  public static void main(String args[]) {\n' +
    '    System.out.println("Hello World !");\n' +
    '  }\n' +
    '}\n',
  cursorOffset: 2,
  comments: []
}

Issue #487, opened 2021-06-19
Support for rangeStart/rangeEnd


jhaber:
**Prettier-Java 1.2.0**

```sh
# Options (if any):
--print-width 90
--range-start 19
--range-end 98
```

**Input:**

```java
package com.test;

public class Test {
  public void sayHello() { System.out.println("hello"); }
}
```

**Output:**

```java
package com.test;

public class Test {

  public void sayHello() {
    System.out.println("hello");
  }
}
```

**Expected behavior:**

It seems like passing range start/end cause nothing to get formatted, even though the code that should get reformatted is within the specified range. Is this just something that isn't implemented yet? 

For some context, I was trying to speed up formatting within our IntelliJ integration by only formatting modified lines  (although maybe it won't make a huge difference if we still need to parse the whole file)

clementdessoude:
This functionality is indeed not implemented, as I was completely unaware of its existence in Prettier !

PR #731, opened 2025-04-27
refactor: leverage Prettier's AstPath and comment attachment

### LINKED ISSUE #534: Wrap before binary operators
Binary operators are being wrapped in a Java-uncommon way.

Binary operator wrapping/indentation was, as far as I can see, last overhauled in https://github.com/jhipster/prettier-java/pull/255 which "fixed" three issues. However, the discussion in https://github.com/jhipster/prettier-java/issues/218 was sadly ignored. The ticket predominantly says that long lines should be wrapped *before* the binary operators, as that's common in Java. And yet the implementation wraps after the operators.

**Prettier-Java 1.6.1**

```sh
# Options (if any):
--print-width 80
```

**Input:**

```java
        String path = "/" + someotherlongstuff + "/";
```

**Output:**

```java
        String path =
            "/" +
            someotherlongstuff +
            "/";
```

**Expected behavior:**
```java
        String path
            = "/"
            + someotherlongstuff
            + "/";
```

or maybe 

```java
        String path = "/"
            + someotherlongstuff
            + "/";
```

This is accented by the fact that for multi-catch we already wrap by "my" rules:

```java
        } catch (
            IOException
            | FTPException
            | NumberFormatException
            | ParseException e
        ) {
```

I understand that the decision has been done at some point and is not very likely to change now as there's already lots of formatted code out there. That said, the discussion around the matter was very clear that Java does things differently than the result, and I'm here to stand by that. At least I wanted to revisit the decision and either explicitly close it forever, or leave it open, gather feedback, and maybe change it in a future major version.

### LINKED ISSUE #592: Formatting not stable with comment in if statement

### LINKED ISSUE #720: wildcard type variable surrounded by line breaks
**Prettier-Java latest main**

Test case in #719

**Input:**

```java
for(SomeClass<?> elem : elements) for(SomeClass<?> elem : elements) for(SomeClass<?> elem : elements)
            doSomeThing();
```

**Output:**

```java
for (SomeClass<?> elem : elements) for (SomeClass<
      ?
    > elem : elements) for (SomeClass<?> elem : elements) doSomeThing();
```

**Expected behavior:**

```java
for (SomeClass<?> elem : elements)
   for (SomeClass<?> elem : elements) 
      for (SomeClass<?> elem : elements)
         doSomeThing();
```

jtkiesel:
## What changed with this PR:

The printer is totally refactored to leverage Prettier's own AstPath and comment attachment to reduce the amount of work done by our own code (including a reduction of ~1700 lines of code). Also improved type safety a bit, specifically by making `CstElement` a union type of `IToken` and all possible `CstNode` types, and in general aligned formatting a bit closer to Prettier's own formatting for JavaScript/TypeScript (one of those happens to be #720).

Because I wanted to prove that it would be easy to resolve issues with comments after refactoring to use Prettier's built-in comment attachment, I also integrated the implementation for #534, while fixing the comment issues I was having in #716.

Because this PR directly leverages Prettier's printing library, there are a few other issues that I have not yet verified, but may be able to be closed with the merge of this PR, including #487 and #660.

Because I renamed some files from `*.js` to `*.ts`, the full diff of the commits in this PR makes it appear as though the `*.js` files were deleted, but I renamed them in the 1st commit to preserve their git history. Because of this, we should avoid squashing this PR's commits when merging, if we want to keep that history.

Something to consider: I removed the java-parser package's custom comment attachment capability, and instead simply attach all comments to the `comments` field of the root node, which is what Prettier expects to allow it to move the comments to the appropriate nodes itself. This would be considered a breaking change to the java-parser package, and so should probably result in a major version increment, if we were to keep this change (though I don't really know how we've been treating breaking changes to the parser in the past). If instead, we determine that we wish to continue supporting our own custom comment attachment in the parser itself, then we could bring it back, and simply make it configurable to disable it. I'm open to either approach, but would obviously prefer to gut the code and reduce our maintenance burden if we can.

### Options

```shell
--experimental-operator-position start
```

### Input

```java
class Example {

  void example() {
    var rotateX =
      (RANGE / rect.height) * refY -
      (RANGE / 2) * getXMultiplication(rect.width);
  }

  void nestedForStatement() {
    for (SomeClass<?> elem : elements) for (SomeClass<
      ?
    > elem : elements) for (SomeClass<?> elem : elements) doSomeThing();
  }
}
```

### Output

```java
class Example {

  void operatorPositionStart() {
    var rotateX =
      (RANGE / rect.height) * refY
      - (RANGE / 2) * getXMultiplication(rect.width);
  }

  void nestedForStatement() {
    for (SomeClass<?> elem : elements)
      for (SomeClass<?> elem : elements)
        for (SomeClass<?> elem : elements) doSomeThing();
  }
}
```

## Relative issues or prs:

Closes #534
Closes #592
Closes #720
Obsoletes #595
Obsoletes #716
Obsoletes #719


packages/prettier-plugin-java/src/printers/blocks-and-statements.ts
```diff
@@ -1,618 +1,495 @@
...
-        localVariableType,
-        variableDeclaratorList
-      ])
-    ]);
-  }
-
-  localVariableType(ctx: LocalVariableTypeCtx) {
-    if (ctx.unannType) {
-      return this.visitSingle(ctx);
+    if (!statements?.length) {
```
clementdessoude, line 28: Could we use the printBody helper here ?

packages/prettier-plugin-java/src/printers/helpers.ts
```diff
@@ -0,0 +1,408 @@
...
+  "Transient",
+  "Volatile",
+  "Synchronized",
+  "Native",
+  "Sealed",
+  "NonSealed",
+  "Strictfp"
+].reduce((map, name, index) => map.set(name, index), new Map<string, number>());
+
+export function sortModifiers(
```
clementdessoude, line 58: I find it strange that this method returns a number and not a list of sorted modifiers. Can we find a more explicit name ?

packages/prettier-plugin-java/src/printers/blocks-and-statements.ts
```diff
@@ -1,618 +1,495 @@
...
+  ifStatement(path, print) {
+    const { children } = path.node;
+    const hasEmptyStatement = isEmptyStatement(children.statement[0]);
+    const statement: Doc[] = [
+      "if ",
+      group([
+        "(",
+        indent([softline, call(path, print, "expression")]),
+        softline,
+        ")"
```
clementdessoude, line 121: Should we have a similar method as putIntoBraces for this kind of code ?

```ts
putIntoBraces(
  path: AstPath<Node>,
  callback: MapCallback<IndexValue<IndexValue<Node, "children">, Child>, U>,
  child: Child
) {
   return group([
        "(",
        indent([softline, call(path, print, "expression")]),
        softline,
        ")"
      ])
}
...
putIntoBraces(path, print, "expression")
```

packages/prettier-plugin-java/src/printers/blocks-and-statements.ts
```diff
@@ -1,618 +1,495 @@
...
+  },
+
+  switchBlock(path, print) {
+    const { children } = path.node;
+    if (!children.switchBlockStatementGroup && !children.switchRule) {
+      const danglingComments = printDanglingComments(path);
+      return danglingComments.length
+        ? ["{", indent([hardline, ...danglingComments]), hardline, "}"]
+        : "{}";
+    }
```
clementdessoude, line 169: Could we not use/mutualize some of the printBlock helper logic ?

packages/prettier-plugin-java/src/printers/blocks-and-statements.ts
```diff
@@ -1,618 +1,495 @@
...
+    if (children.Semicolon) {
+      parts.push(";");
+    }
+    return parts;
+  },
+
+  caseConstant: printSingle,
+  casePattern: printSingle,
+
+  whileStatement(path, print) {
```
clementdessoude, line 254: I see that there is no handling of dangling comment in the while loop, is it because it cannot happen ? Or handled by default ? Or something we need to handle afterwards ?

packages/prettier-plugin-java/src/printers/blocks-and-statements.ts
```diff
@@ -1,618 +1,495 @@
...
-          concat([ctx.RBrace[0], " "])
-        ])
-      ),
-      block
-    ]);
-  }
+  resourceSpecification(path, print, options) {
+    const resources = [call(path, print, "resourceList")];
+    if (options.trailingComma !== "none") {
+      resources.push(ifBreak(";"));
```
clementdessoude, line 479: Could this not lead to a double ";" if there is already a trailing ";" ?

clementdessoude (commented):
Hello @jtkiesel ! Great work also here. It seems that it will certainly lessen the maintaining burden and ease comments handling.

I didn't had the time to fully review the MR this morning (there are quite some changes). I think I should be able to continue (and hopefully finish it) before Tuesday evening. 

Let me know if you do not agree with some of the comments, we can discuss them ;)

packages/prettier-plugin-java/src/printer.ts
```diff
@@ -1,7 +1,54 @@
...
 
-// eslint-disable-next-line no-unused-vars
-export default function genericPrint(path, options, print) {
-  const node = path.getValue();
-  return createPrettierDoc(node, options);
+export default {
+  print(path: DistributedAstPath<CstElement>, options, print, args) {
+    return hasToken(path)
+      ? path.node.image
+      : printerForNodeType(path.node.name)(path, print, options, args);
```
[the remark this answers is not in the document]
jtkiesel: It was intended because the `options` argument is often (almost always) unneeded for the printer functions, and I wanted them to all have the same signature, but for `options` and `args` to be optional, since they are rarely used.

packages/prettier-plugin-java/src/printers/helpers.ts
```diff
@@ -0,0 +1,408 @@
...
+  if (node.name === "argumentList") {
+    return locEndWithComments(node.children.expression.at(-1)!);
+  } else if (node.name === "blockStatements") {
+    return locEndWithComments(node.children.blockStatement.at(-1)!);
+  }
+  const { comments } = node;
+  return comments
+    ? Math.max(parser.locEnd(node), parser.locEnd(comments.at(-1)!))
+    : parser.locEnd(node);
+}
```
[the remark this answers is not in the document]
jtkiesel: Thanks for catching this! These functions were used in a `printIgnored` function when I was originally keeping the handling of `@formatter:off`/`@formatter:on` comments in the parser, and so had to handle printing the ignored sections myself. More recently (last week) I pushed a change that integrated the handling of those comments into `hasPrettierIgnore` (via `isFullyBetweenFormatterOffOn`), so now Prettier handles the printing of those ignored sections, so these functions are no longer needed. 🎉

packages/prettier-plugin-java/src/printers/blocks-and-statements.ts
jtkiesel: It won't, because the `resourceList` printer performs a join on the resources with `;` as the delimiter. The trailing `;` would exist on `ResourceSpecification` itself, if it exists at all, and we ignore whether it exists already or not so that the `trailingComma` option is obeyed no matter what.

packages/prettier-plugin-java/src/printers/blocks-and-statements.ts
jtkiesel: Correct, for these statements (`while`, `do`, `for`), they cannot have dangling comments because they are required to have a `statement` child, and so Prettier will always be able to attach comments to the statement. The reason we have to handle dangling comments for other node types such as blocks/bodies is because they could consist of only non-Identifier tokens (i.e. `{}`), and so Prettier won't be able to attach comments that exist in the middle of those tokens because it wouldn't be appropriate to attach them as leading or trailing (if they happened to be before or after all tokens, it would attach them as leading/trailing comments to the block/body node itself), so it is forced to attach them as dangling comments (it knows they should be somewhere in the middle of the tokens, but requires us to handle the printing of them).

packages/prettier-plugin-java/src/printers/helpers.ts
```diff
@@ -0,0 +1,408 @@
...
+} from "java-parser";
+import type { AstPath, Doc, ParserOptions } from "prettier";
+import { builders } from "prettier/doc";
+import parser from "../parser.js";
+
+const { group, hardline, indent, join, line } = builders;
+
+export function onlyDefinedKey<
+  T extends Record<string, any>,
+  U extends Key<T> & string
```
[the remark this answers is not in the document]
jtkiesel: Trying to determine a good balance between more explicit naming and still making it clear that these are generics and not simply other types (I already use `Key` as a type name, for example, so I can't use that here even if I wanted).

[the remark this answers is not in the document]
jtkiesel: I could align with Prettier, where I copied a number of these generic type definitions from? They seem to use `T` for "input" and `U` for "output", and `P` for "prop". TypeScript seems to use a similar convention, they'll also use `K` for "key".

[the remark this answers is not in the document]
jtkiesel: Decided to align with Prettier's/TypeScript's conventions for now, let me know your thoughts.

packages/prettier-plugin-java/src/printers/helpers.ts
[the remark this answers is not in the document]
jtkiesel: Removed these.

packages/prettier-plugin-java/src/printers/blocks-and-statements.ts
jtkiesel: Good idea! I renamed `printBody` to `printBlock`, and used it here and in a number of other places to reduce duplication.

packages/prettier-plugin-java/src/printers/helpers.ts
jtkiesel: Good point, I meant to come back to this and rename this function, but forgot. I don't love the wordiness of `sortModifiersAndGetDeclarationAnnotationCount`, but that's what I went with for now.

packages/prettier-plugin-java/src/printers/blocks-and-statements.ts
jtkiesel: Good idea! I created `indentInParentheses` for this purpose and used in everywhere I could.

jtkiesel:
> Hello @jtkiesel ! Great work also here. It seems that it will certainly lessen the maintaining burden and ease comments handling.
> 
> I didn't had the time to fully review the MR this morning (there are quite some changes). I think I should be able to continue (and hopefully finish it) before Tuesday evening.
> 
> Let me know if you do not agree with some of the comments, we can discuss them ;)

Thank you for taking the time to review, @clementdessoude! I've resolved/responded to all your comments, happy to continue discussing as well. I'll look forward to the rest of your review! Thanks again for taking the time, I know it's a large PR. 😅

jtkiesel:
Just made some minor changes to improve the types on both the parser and the printer. Made it clear via the types that only the root node from the parser now contains comments, and added comments to the types on the printer side (where Prettier is now attaching them to the appropriate nodes). Also renamed some types/functions in the printer, because "node" vs "token" was a bit confusing due to Prettier using "node" to describe all elements of the AST, so I went with "non-terminal" and "terminal".

packages/prettier-plugin-java/test/unit-test/expressions/_output.java
```diff
@@ -194,8 +194,7 @@ public void printSynchronized() {
   }
 
   public void longFullyQualifiedName() {
-    com.me.very.very.very.very.very.very.very.very.very.very.very.very.very.longg.fully.qualified.name.FullyQualifiedName.builder()
-      .build();
+    com.me.very.very.very.very.very.very.very.very.very.very.very.very.very.longg.fully.qualified.name.FullyQualifiedName.builder().build();
```
clementdessoude, line 197: Is this intended or should it break like before ?

packages/prettier-plugin-java/test/unit-test/template-expression/_output.java
```diff
@@ -6,11 +6,11 @@ class TemplateExpression {
 
   String s = STR."You have a \{getOfferType()} waiting for you!";
 
-  String msg =
-    STR."The file \{filePath} \{file.exists() ? "does" : "does not"} exist";
+  String msg = STR."The file \{filePath} \{
```
clementdessoude, line 9: I think we could remove String template handling in the future, if it becomes difficult to handle, as it has been removed from the spec (but not the purpose of this PR ofc !)

packages/prettier-plugin-java/test/unit-test/text-blocks/_output.java
```diff
@@ -1,30 +1,27 @@
...
 
-    String html =
-      """
-      <html>\r
-          <body>\r
-              <p>Hello, world</p>\r
-          </body>\r
-      </html>\r
-      """;
+    String html = """
```
clementdessoude, line 18: The indentation seems a bit strange here, don't you think ?

packages/prettier-plugin-java/test/unit-test/template-expression/_output.java
jtkiesel: Yeah, I nearly decided to remove it in this PR because trying to fix its formatting was becoming difficult. As long as we're okay with the way the formatting of them is in this PR (which I was, since it was a preview feature to begin with and has since been removed entirely, as you pointed out), then perhaps we keep this as-is and remove template expressions in a subsequent PR prior to the creation of our next release.

packages/prettier-plugin-java/test/unit-test/expressions/_output.java
jtkiesel: This change was kept intentionally, as it aligns with Prettier's JS/TS formatting ([example](https://prettier.io/playground/#N4Igxg9gdgLgprEAucAbAhgZ0wAgKIAe6AtgA6pw7AA6UOOcRZFAFAJRW332TFff0AdMTiCAbnABOAT3FTZEmXKWKF85WpXrVG3TtTQA5ocEAzAK6pUsgI7n0qAJanHcACaCoJUQDFL1gEV7Jxd3ADlvQQAjc0dUNyl2fgFo2Pj2AG5+AF9aXKgQABoQCFIYR2hMZFB0SUkIAHcABVqEKpQHBvRpKuKoyXQwAGs4GABlbwAZRyg4ZFMHTDhiiCiAKzgwGAB1AdJkEFJJOCXJCSKQfsGR8dJBmcNkGElzZZAl4kcnl7fGUilHCJYA4APL-AYwCCSJoQTCOcrQA4INwXP4AoEwBwAFSkXkkrnaC1QS2KcKghgoQQg8Hmizea0wBDGD0p5mpcyQRJJIDs7Ka9VI7RAWAAtLN3O4Ls90HEHgBhCDEYjoA4OVAXMkUuAAQRgz0cMXgTSk01mtOJbwAFjBiKhtpb4Sc7mA4GM2vDHGJ4dIDmBsBcxK8AJJQBKwMZgfFlbWhsYwaQUc3co6wuC7dD7FBHE5Sc7FGanGD89CGZVJt53SSnA7x-6YSOOMoXI4zHaONwwS3IAAcAAZisc7I5jsXSyrOXTipiott253kAAmYrmJZY9BRQmTkBwYhRdwJNyTdDk+yGOA+KHKvUPVXmSEgbLZIA)).

packages/prettier-plugin-java/test/unit-test/text-blocks/_output.java
clementdessoude: Putting the opening """ at the right of the equal is fine by me. It is the indentation below that seems a bit odd, I would add one indent as it was before.
It differs a bit from JS/TS as the leading spaces are not removed in JS/TS as it is in Java with the closing """.

packages/prettier-plugin-java/src/printers/classes.ts
```diff
@@ -1,1074 +1,703 @@
...
+    const declarators = map(path, print, "variableDeclarator");
+    return declarators.length > 1 &&
+      path.node.children.variableDeclarator.some(
+        ({ children }) => children.Equals
+      )
+      ? group(indent(join([",", line], declarators)), {
+          shouldBreak:
+            (path.getNode(4) as JavaNonTerminal | null)?.name !== "forInit"
+        })
+      : join(", ", declarators);
```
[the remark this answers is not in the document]
jtkiesel: Hmm, does seem to be a new use case, and I'm not sure how/why I ran into it without there being any existing tests for it. [This](https://prettier.io/playground/#N4Igxg9gdgLgprEAucAbAhgZ0wAgKIAe6AtgA6pw7AA6UOOcRZFAFAJRW33UwUw7oANDgBGwsDgC8OAIwBuLjgC+tFVBCCQEUjACW0TMlDoATiYgB3AAqmEhlOlQX0AT0OaRJ9GADWcGADKJHAAMrpQcMgAZo6YcJoQIgBWcGAwAOpepMggpCZwcSYAbpEeXr7+AaTe4QDmyDAmAK7xIHHEug3NrYykcCa6xAgwjgDyfV4wECZWEJi6etA5CAAmGiC9-YPDjgAq-VCmugXRsa3zULUUAIpNEPCnqHGaSZgEAXU3dw9IMU+tAEdvnArOZSPYQFgALQROArOHrRroXSoOoAYQgxGI6ByjlQ6wuVzgAEEYI1dCImvArP0whFHs8QAALGDEVDpJkLArVMBwAJ2Ba6IoLFw5MDYdZFFoASSg8NgATAAx0xLlARgLgoDNaeTmcEy6GyKDyBX6JXW4UKMFB6Fq2O1mmqJkKOQ1fUwSt0OnWeXCGV0KxgTOQAA4AAyafJA3T5G12nG-M6aEYidIBoPIABMmiacV26BE9j+jLgxBEcPhKxC6EuTVtcAAYtNsWS6riqRAQEolEA) is an example of the use case being handled here. I've now added a unit test for it [here](https://github.com/jhipster/prettier-java/blob/refactor/prettier-path-and-comment/packages/prettier-plugin-java/test/unit-test/variables/_output.java#L365).

packages/prettier-plugin-java/src/printers/classes.ts
```diff
@@ -1,1074 +1,703 @@
...
+    }
+    return declarator;
+  },
 
-    return rejectAndJoin(" ", [
+  receiverParameter(path, print) {
+    const { children } = path.node;
+    const annotations = children.annotation
+      ? map(path, print, "annotation")
+      : [];
```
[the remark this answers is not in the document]
jtkiesel: Prettier's `path.map` fails if the value of the key on the node is `undefined`, but I went ahead and just baked this handling into the regular `map` helper, similarly to the way you suggested.

packages/prettier-plugin-java/src/printers/classes.ts
```diff
@@ -1,1074 +1,703 @@
...
-
-    const blankLineSeparators = getBlankLinesSeparator(ctx.enumConstant);
-    const commas = ctx.Comma
-      ? ctx.Comma.map((elt, index) =>
-          concat([elt, blankLineSeparators![index]])
-        )
+  },
+
+  enumConstant(path, print) {
+    const { argumentList, classBody, enumConstantModifier } =
```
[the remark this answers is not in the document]
jtkiesel: Great point! I ended up being able to get rid of the ugly-named `sortModifiersAndGetDeclarationAnnotationCount` and replace it with `printWithModifiers`. The old function also had to modify the CST in-place to sort the modifiers, so this new way is much better.

packages/prettier-plugin-java/test/unit-test/binary_expressions/operator-position-end/_output.java
```diff
@@ -85,8 +85,7 @@ public void method() {
     );
 
     // Issue 381
-    new MethodWrappingFollowingContstructor()
-      .aLongEnoughMethodNameToForceThingsToWrap();
+    new MethodWrappingFollowingContstructor().aLongEnoughMethodNameToForceThingsToWrap();
```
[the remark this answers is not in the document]
jtkiesel: This change was kept intentionally, as it aligns with Prettier's JS/TS formatting ([example](https://prettier.io/playground/#N4Igxg9gdgLgprEAucAbAhgZ0wAgKIAe6AtgA6pw7AA6UOOcRZFAFAJRW331RwDuOALJwYACwgATAGIRUqCHwCWUAOYBhaDEwwATgFcwMCDvYA6dABloKvFAh6Vo4WMkA5EnAAqEQvCgSABSwYAJ1lGAB1RQkxdgBuLhwAX1oUqBAAGhAIUhhFaExkUHQdHQUgnQRClHRUPnQAT0KsgCMddDAAaxEAZQ8LZThkADNazDgsiBaAKzhDCPbSZBBSSvGdADch1vau3tIO5RVkXT0JkHHiRRP9c8ZSODDiBBhagHkH9qMdAIhMRTy0GWCAkmRA90eimesFqnkeUBKijg1VGqHGWX+qgoAEU9BB4CMxudppgCD0jji8QSkKj0SAAI5UuChHLVEBYAC0vDgEh5YN06EUqCOGmIxHQy1qqDBmJUFAAgjBdIoWnp4AFHgNeIS0edRDBiKgIqIAciDmA4D0qgDFBsAQ1lmBsGCNmcAJL+F49MBhXLy-w9GANCg6umrP5wBboJYoVbIx5bMHKdYhdoqcWh84HHTrZZBh6YH2KXJg1bhKIxUTIAAcAAYspVGYpKqF0OmJTSiVlXi0K2JkAAmLJ6caedAtFFd8HEFo83kSCzoVR6NtwGQ6cVKo6StUQEBJJJAA)). I've updated the class/method names to be more appropriate, and removed the comment.

packages/prettier-plugin-java/test/unit-test/text-blocks/_output.java
jtkiesel: That's fair. That does seem to be the most common formatting in Java's own examples (opening `"""` on the same line as the equals, with the rest indented by 1 level): https://docs.oracle.com/en/java/javase/15/text-blocks/index.html

I've made this change now, so that the indentation is the same as it was previously.

packages/prettier-plugin-java/src/printers/expressions.ts
```diff
@@ -1,901 +1,784 @@
...
+    const keyword = call(path, print, keywordKey);
+    return children.typeArguments
+      ? [keyword, call(path, print, "typeArguments")]
+      : keyword;
+  },
+
+  parenthesisExpression(path, print) {
+    const expression = call(path, print, "expression");
+    const ancestorName = (path.getNode(14) as JavaNonTerminal | null)?.name;
+    const binaryExpression = path.getNode(8) as JavaNonTerminal | null;
```
clementdessoude, line 358: what are those path.getNode(14) and path.getNode(8) ?

clementdessoude (commented):
Hey @jtkiesel ! Finally had some time to finish reviewing the PR ! I have to admit that there are some areas of the code where I didn't go in complete details but I trust the unit test enough to be ok with it 👍 

Just have some questions about the `path.getNode` but otherwise looks good to me !

packages/prettier-plugin-java/src/printers/lexical-structure.ts
```diff
@@ -1,45 +1,40 @@
...
-    }
-    return this.visitSingle(ctx);
-  }
+    const [open, ...lines] = TextBlock[0].image.split("\n");
+    const baseIndent = findBaseIndent(lines);
+    const textBlock = join(hardline, [
+      open,
+      ...lines.map(line => line.slice(baseIndent))
+    ]);
+    const ancestor = path.getNode(14) as JavaNonTerminal | null;
```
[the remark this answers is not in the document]
jtkiesel: `.getNode()` allows you to obtain an ancestor node by index up the stack. So this is 14 parents up, which because of the `.children` field on every node means it's actually 7 nodes up the stack. The `parent` and `grandparent` values on `AstPath` use this function internally. Here we're obtaining this specific ancestor in order to check if this TextBlock literal is a descendant of a VariableInitializer or BinaryExpression with an assignment operator.

packages/prettier-plugin-java/src/printers/expressions.ts
jtkiesel: See https://github.com/jhipster/prettier-java/pull/731#discussion_r2141655318 for generic explanation. In this case we're obtaining these specific ancestors (7 and 4 nodes up the stack) to check if this parentheses expression is the descendent of a BinaryExpression in a Guard or ReturnStatement.

murdos:
Hi @jtkiesel,
Do you plan to do a release with the new features that this PR is bringing? I was thinking e.g. to #534
Thanks!

DanielFran:
> @DanielFran Bounty claimed: https://opencollective.com/generator-jhipster/expenses/254462

@jtkiesel approved

PR #741, opened 2025-07-03
Fix update-test-outputs script to use test-specific prettier options.

motlin:
Before this change, checking out the project and running update-test-outputs would make changes to the test outputs.

```
❯ npm run build && npm run update-test-outputs
❯ git status
HEAD detached at upstream/main
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   packages/prettier-plugin-java/test/unit-test/binary_expressions/operator-position-start/_output.java
        modified:   packages/prettier-plugin-java/test/unit-test/require-pragma/invalid-pragma/_output.java

no changes added to commit (use "git add" and/or "git commit -a")
```

The change here is to use the test-specific configuration while recording tests. With this change, running npm run update-test-outputs doesn't make any changes to any files.

### DIFF
--- a/packages/prettier-plugin-java/scripts/update-test-output.js
+++ b/packages/prettier-plugin-java/scripts/update-test-output.js
@@ -51,12 +51,20 @@ const updateTestOutput = async () => {
       try {
         console.log(`Reading <${fileDesc.path}>`);
         let newExpectedText = javaFileText;
+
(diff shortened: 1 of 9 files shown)


jtkiesel (approved):
Thank you for the quick changes and the rebase! It looks great. 👍

Issue #746, opened 2025-07-10
prettier-plugin-java types are broken.


mshima:
prettier-plugin-java types are using relative paths to reference java-parser.
```js
    printers: {
        java: {
            print(path: import("prettier").AstPath<import("../../java-parser/api.js").ArrayInitializerCstNode & {
                comments?: import("./comments.js").JavaComment[];
            }> | import("prettier").AstPath<import("../../java-parser/api.js").VariableInitializerListCstNode & {
```

It fails to find types if java-parser is installed deeper in node_modules path.

See https://github.com/jhipster/generator-jhipster/pull/29952

PR #747, opened 2025-07-10
chore: fix generated types

mshima:
## What changed with this PR:

Do not generate relative path in generated types.

## Relative issues or prs:

Fixes https://github.com/jhipster/prettier-java/issues/746

### DIFF
--- a/packages/prettier-plugin-java/tsconfig.json
+++ b/packages/prettier-plugin-java/tsconfig.json
@@ -7,10 +7,7 @@
     "esModuleInterop": true,
     "noImplicitThis": false,
     "module": "ESNext",
-    "moduleResolution": "Node",
-    "paths": {
-      "java-parser": ["../java-parser"]
-    }
+    "moduleResolution": "Node"
   },
   "include": ["src"]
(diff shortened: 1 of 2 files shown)


packages/prettier-plugin-java/tsconfig.json
```diff
@@ -7,10 +7,7 @@
     "esModuleInterop": true,
     "noImplicitThis": false,
     "module": "ESNext",
-    "moduleResolution": "Node",
-    "paths": {
-      "java-parser": ["../java-parser"]
```
jtkiesel, line 12: I thought this was added so that the local java-parser was used when running prettier-plugin-java tests, or to ensure the local java-parser types were used when developing prettier-plugin-java, but removing this doesn't seem to affect that, so I'm not sure why it was added. It seems like it was added unnecessarily, and I'm confident it was added for development purposes only, so even if there was a good reason for it, we'll just need to find a way that doesn't break the packaged types.

jtkiesel (approved):
Thanks for finding and fixing this!

PR #840, opened 2026-04-05
Add braceStyle option with same-line (default) and next-line choices.

motlin:
## What changed with this PR:

Added a `braceStyle` option with two choices:
- `same-line` (default) — opening braces on the same line (K&R style, existing behavior)
- `next-line` — opening braces on the next line (Allman style)

This applies to all block constructs: class/interface/enum bodies, method bodies, control flow statements (if/else, for, while, do-while, try/catch/finally, switch, synchronized), static and instance initializers, anonymous classes, lambda blocks, and array initializers.

### Input

```java
public class HelloWorld {
    public void greet(String name) {
        if (name != null) {
            System.out.println("Hello, " + name);
        } else {
            System.out.println("Hello, World");
        }
    }
}
```

### Output

With `{ "braceStyle": "next-line" }`:

```java
public class HelloWorld
{
  public void greet(String name)
  {
    if (name != null)
    {
      System.out.println("Hello, " + name);
    }
    else
    {
      System.out.println("Hello, World");
    }
  }
}
```

## Relative issues or prs:

- [prettier/prettier#4212 — Option for brace style](https://github.com/prettier/prettier/issues/4212)
- [prettier/prettier#3964 — Is it possible to setup "allman" braces style?](https://github.com/prettier/prettier/issues/3964)
- [prettier/prettier#3832 — Add option to set braces style](https://github.com/prettier/prettier/issues/3832)

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -31,5 +31,21 @@ export default {
       { value: "end", description: "" }
     ],
     description: "Where to print operators when binary expressions wrap lines."
+  },
+  braceStyle: {
+    type: "choice",
+    category: "Java",
+    default: "same-line",
+    choices: [
+      {
+        value: "same-line",
+        description: "Put opening braces on same line (K&R style)"
+      },
+      {
+        value: "next-line",
+        description: "Put opening braces on new lines (Allman style)"
+      }
+    ],
+    description: "Placement of opening braces."
   }
 } satisfies SupportOptions;
--- a/src/printers/blocks-and-statements.ts
+++ b/src/printers/blocks-and-statements.ts
@@ -25,8 +25,8 @@ const {
 } = builders;
 
 export default {
-  block(path, print) {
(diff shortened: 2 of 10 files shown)


jtkiesel:
@motlin Thank you for your interest in contributing to Prettier Java!

Prettier is an opinionated formatter, and this Java plugin is as well. Prettier themselves have decided against supporting other brace styles (as is evidenced by every issue you linked being closed), and the K&R style is the [standard](https://en.wikipedia.org/wiki/Indentation_style#Java) and dominant style for braces in Java. Prettier offers very few options on [purpose](https://prettier.io/docs/option-philosophy), and almost never add new options as it goes directly against the primary goal of Prettier (to avoid debates over different styles). As such, we will not integrate this option.

That being said, if you have other ideas for improvements to Prettier Java, please do create an issue as I am more than happy to discuss their viability and whether we would accept such a contribution!

PR #892, opened 2026-05-18
fix: correct Windows path resolution in parser init

ralscha:
## What changed with this PR:

Fix the invalid path on Windows when loading the treesitter wasm file

## Example
```
 no such file or directory, open 'C:\C:\.....\target\spotless-prettier-node-modules-a419b185f88f4c8243af8729ce04d5c8\node_modules\prettier-plugin-java\dist\tree-sitter-java_orchard.wasm'
```

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1,4 +1,3 @@
-import { fileURLToPath } from "node:url";
 import type Prettier from "prettier";
 import { Language, Parser, type Node } from "web-tree-sitter";
 import { determinePrettierIgnoreRanges } from "./comments.ts";
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -42,7 +42,7 @@ const parser = (async () => {
 
   const parser = new Parser();
   const Java = await Language.load(
-    fileURLToPath(new URL("./tree-sitter-java_orchard.wasm", import.meta.url))
+    new URL("./tree-sitter-java_orchard.wasm", import.meta.url)
   );
   parser.setLanguage(Java);
 
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1,3 +1,4 @@
+import { fileURLToPath } from "node:url";
 import type Prettier from "prettier";
 import { Language, Parser, type Node } from "web-tree-sitter";
 import { determinePrettierIgnoreRanges } from "./comments.ts";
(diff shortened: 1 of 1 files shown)


jtkiesel:
Thanks for the contribution (and inherently reporting this issue)! Unfortunately, this breaks use of this plugin in web (as can be seen by the failed GitHub Pages CI job). We purposefully omitted use of Node.js-only libraries for this reason. I will see if there's another way we can resolve this issue.

src/parser.ts
```diff
@@ -41,7 +42,7 @@ const parser = (async () => {
 
   const parser = new Parser();
   const Java = await Language.load(
-    new URL("./tree-sitter-java_orchard.wasm", import.meta.url).pathname
+    fileURLToPath(new URL("./tree-sitter-java_orchard.wasm", import.meta.url))
```
jtkiesel: Can you try this out instead?

```suggestion
    new URL("./tree-sitter-java_orchard.wasm", import.meta.url)
```

Even though this fails type checking, it seems to work (I'm assuming tree-sitter-web just passes the argument directly to `fs.readFile()` or similar). I don't have an easy way to test this in Windows, but assuming you do, your help would be greatly appreciated!

ralscha: This works on Windows

```
import { expect } from "chai";
import { Language, Parser } from "web-tree-sitter";

describe("wasm-path-resolution", () => {
  it("Language.load accepts a URL object to the wasm file", async () => {
    await Parser.init();

    const wasmUrl = new URL(
      "../../../dist/tree-sitter-java_orchard.wasm",
      import.meta.url
    );

    const Java = await Language.load(wasmUrl);
    expect(Java).to.be.instanceOf(Language);
    expect(wasmUrl.protocol).to.equal("file:");

    if (process.platform === "win32") {
      expect(wasmUrl.pathname).to.match(/^\/[A-Za-z]:\//);
      expect(wasmUrl.href).to.match(/^file:\/\/\/[A-Za-z]:\//);
    } else {
      expect(wasmUrl.pathname).to.match(/^\//);
    }
  });
});
```

ralscha: Here is the code of Language.load. You are right it uses `fs.readFile`

```
  static async load(input: string | Uint8Array): Promise<Language> {
    let binary: Uint8Array | WebAssembly.Module;
    if (input instanceof Uint8Array) {
      binary = input;
    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
    } else if (globalThis.process?.versions.node) {
      const fs: typeof import('fs/promises') = await import('fs/promises');
      binary = await fs.readFile(input);
    } else {
      const response = await fetch(input);
```

jtkiesel: Excellent! Let's go with that then.

src/parser.ts
```diff
@@ -1,3 +1,4 @@
+import { fileURLToPath } from "node:url";
```
jtkiesel:
```suggestion
```

